PR #3, opened 2016-08-29
add jaeger.thrift

oibe:


thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+# THE SOFTWARE.
+
+namespace java com.uber.jaeger.thriftjava
+
+# TagType denotes the type of a Tag's value.
+enum TagType { STRING, DOUBLE, BOOL, I16, I32, I64, BINARY }
+
+# Tag is a basic strongly typed key/value pair. It has been flattened to reduce the use of pointers in golang
+struct Tag {
+  1: required string   key
```
yurishkuro: space

thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+}
+
+# Batch is a collection of spans reported out of process.
+struct Batch {
+  1: required Process    process
+  2: required list<Span> spans
+}
+
+service Agent {
+    oneway void emitJaegerBatch(1: list<Span> spans)
```
yurishkuro: rm Jaeger

thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+struct Tag {
+  1: required string   key
+  2: required TagType tagType
+  3: required string  vStr
+  4: required double  vDouble
+  5: required bool    vBool
+  6: required i16     vInt16
+  7: required i32     vInt32
+  8: required i64     vInt64
+  9: required binary  vBinary
```
yurishkuro: I am uncomfortable committing to `required` without a benchmark. Especially in non-compact encoding, how much large will the entry be on the wire?

thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+}
+
+# Log is a timed even with an arbitrary set of tags.
+struct Log {
+  1: required i64       timestamp
+  2: required list<Tag> tags
```
badiib: there's some inconsistency with respect to tags either being required or not in the span, process, and log object.

thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+  3: required i64         spanId
+}
+
+# Span represents a named unit of work performed by a service.
+struct Span {
+  1: required i64           traceId      # unique trace id, the same for all spans in the trace
+  2: required i64           spanId       # unique span id (only unique within a given trace)
+  3: required string        operationName
+  4: optional list<SpanRef> references    # causal references to other spans
+  5: required i32           flags         # tbd
```
badiib: required vs optional here too. Will it always be available?

yurishkuro: flags are required

thrift/jaeger.thrift
yurishkuro: hm, the semantics of required vs. optional list is not very clear. In case of a Log, the list is not only required, but also has to be non-empty. In a Span the list of tags could be empty, as can be the list of references. Keeping those lists optional avoids memory allocation of an empty list, so I think the current qualifiers are correct.

thrift/jaeger.thrift
```diff
@@ -0,0 +1,81 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+namespace java com.uber.jaeger.thriftjava
```
yurishkuro: s/thriftjava/thrift

yurishkuro:
LGTM

PR #6, opened 2016-09-01
making a parent span ID a first class citizen as it will exist nearly…

badiib:
… all the time. reducing span ID to 32 bits as 64 bits is overkill

### DIFF
--- a/thrift/jaeger.thrift
+++ b/thrift/jaeger.thrift
@@ -47,22 +47,22 @@ struct SpanRef {
   1: required SpanRefType refType
   2: required i64         traceIdLow
   3: required i64         traceIdHigh
-  4: required i64         spanId
+  4: required i32         spanId
 }
 
 # Span represents a named unit of work performed by a service.
 struct Span {
-  1:  required i64           traceIdLow   # the last 64 bits of a traceID
-  2:  required i64           traceIdHigh  # the first 64 bits of a traceID
+  1:  required i64           traceIdLow   # the least significant 64 bits of a traceID
+  2:  required i64           traceIdHigh  # the most significant 64 bits of a traceID; 0 when only 64bit IDs are used
   3:  required i32           spanId       # unique span id (only unique within a given trace)
-  4:  required string        operationName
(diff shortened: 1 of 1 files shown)


thrift/jaeger.thrift
```diff
@@ -54,14 +54,15 @@ struct SpanRef {
...
   4:  required string        operationName
-  5:  optional list<SpanRef> references    # causal references to other spans
-  6:  required i32           flags         # tbd
+  5:  optional list<SpanRef> references   # causal references to other spans
+  6:  required i32           flags        # tbd
   7:  required i64           startTime
   8:  required i64           duration
   9:  optional list<Tag>     tags
   10: optional list<Log>     logs
+  11: optional i32           parentId     # since nearly all spans will have parents spans, CHILD_OF refs do not have to be explicit
```
yurishkuro: a) I would move it to position 4
b) I am not sure it needs to be optional, because 0 is the natural "not present" value here

thrift/jaeger.thrift
```diff
@@ -54,14 +54,15 @@ struct SpanRef {
 struct Span {
   1:  required i64           traceIdLow   # the last 64 bits of a traceID
   2:  required i64           traceIdHigh  # the first 64 bits of a traceID
-  3:  required i64           spanId       # unique span id (only unique within a given trace)
+  3:  required i32           spanId       # unique span id (only unique within a given trace)
```
yurishkuro: I assume you also need to change the Reference type

thrift/jaeger.thrift
```diff
@@ -54,14 +54,15 @@ struct SpanRef {
 struct Span {
   1:  required i64           traceIdLow   # the last 64 bits of a traceID
   2:  required i64           traceIdHigh  # the first 64 bits of a traceID
```
yurishkuro: * in the comments, replace last/first with "least/most significant"
* add to High comment: "; 0 when only 64bit IDs are used"

yurishkuro:
LGTM

PR #8, opened 2016-09-14
update SpanRef spanId to int64

black-adder:


badiib:
LGTM, pending checks.

PR #11, opened 2016-10-24
Add OperationSamplingStrategy

black-adder:


thrift/sampling.thrift
```diff
@@ -35,10 +35,16 @@ struct RateLimitingSamplingStrategy {
     1: required i16 maxTracesPerSecond
 }
 
+struct OperationSamplingStrategy {
```
vprithvi: nitpick: Could you add a comment on what this struct is used for?

PR #12, opened 2016-10-25
added required to zipkincore.thrift

oibe:


oibe:
@yurishkuro Now that we are moving the tchannel bridge to the new-client we need this.

thrift/zipkincore.thrift
```diff
@@ -164,21 +164,21 @@ struct Endpoint {
    *
    * Ex for the ip 1.2.3.4, it would be (1 << 24) | (2 << 16) | (3 << 8) | 4
    */
-  1: i32 ipv4
+  1: required i32 ipv4
```
badiib: changing this has very far-reaching side effects across multiple languages. What are you trying to accomplish here?

PR #13, opened 2016-10-27
Update operation sampling strategy

black-adder:
This is a breaking change, I might have to keep 4 as is and add 5 on SamplingStrategyResponse if someone has already deployed a client with the old change.

### DIFF
--- a/thrift/sampling.thrift
+++ b/thrift/sampling.thrift
@@ -53,8 +53,7 @@ struct SamplingStrategyResponse {
     1: required SamplingStrategyType strategyType
     2: optional ProbabilisticSamplingStrategy probabilisticSampling
     3: optional RateLimitingSamplingStrategy rateLimitingSampling
-    // 4 is deprecated
-    5: optional OperationSamplingStrategies operationSampling
+    4: optional OperationSamplingStrategies operationSampling
 }
 
 service SamplingManager {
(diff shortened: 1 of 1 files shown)


thrift/sampling.thrift
```diff
@@ -35,17 +35,25 @@ struct RateLimitingSamplingStrategy {
...
+    1: required string operation
+    2: required ProbabilisticSamplingStrategy probabilisticSampling
 }
 
 struct SamplingStrategyResponse {
     1: required SamplingStrategyType strategyType
     2: optional ProbabilisticSamplingStrategy probabilisticSampling
     3: optional RateLimitingSamplingStrategy rateLimitingSampling
-    4: optional list<OperationSamplingStrategy> operationSamplingStrategies
+    5: optional OperationSamplingStrategies operationSampling
```
yurishkuro: do any of our clients already use the previous schema?

you should leave a comment saying field 4: is deprecated

black-adder: I updated the go client to use it a couple of days ago. However, since I can't tell if it's already in the wild, I think the save approach is to use 5.

Will add comment

yurishkuro:
Any other clients besides Go were upgraded to the new IDL? I don't think so.

And the Go client hasn't been released, even though we pushed a commit.

black-adder:
Checked with team, 4 is good to go.

thrift/sampling.thrift
```diff
@@ -35,17 +35,25 @@ struct RateLimitingSamplingStrategy {
     1: required i16 maxTracesPerSecond
 }
 
-// OperationSamplingStrategy is the strategy used per operation.
+// OperationSamplingStrategies samples traces per operation with a guaranteed lower bound per second. Once
```
yurishkuro: * nit: I think "PerOperation" is a more accurate name (here and in field 4 below, but not in L47)
* nit: comment should read "PerOperationSamplingStrategies defines a sampling strategy per each operation name in the service..."

PR #14, opened 2016-10-28
Change thrift ordering

black-adder:
Sooooo, the generated python code doesn't like the ordering.


yurishkuro:
Would #15 have helped to catch it earlier?

PR #16, opened 2016-12-16
adding collector signature for jaeger

badiib:
am i missing something? should i be generating the .go files?

### DIFF
--- a/thrift/jaeger.thrift
+++ b/thrift/jaeger.thrift
@@ -77,6 +77,14 @@ struct Batch {
   2: required list<Span> spans
 }
 
+struct BatchSubmitResponse {
+    1: required bool ok
+}
+
 service Agent {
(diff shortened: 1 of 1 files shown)


thrift/jaeger.thrift
```diff
@@ -77,6 +77,14 @@ struct Batch {
   2: required list<Span> spans
 }
 
+struct BatchSubmitResponse {
+    1: required bool ok
```
yurishkuro: Clarify the semantics. What does "not ok" mean? What is the agent expected to do in that case?

badiib: good question. i haven't thought about it. not-ok is just that it failed.

yurishkuro: I suggest adding a comment that that Agent is not expected to do anything with not-ok response, other than perhaps increment a counter.

PR #19, opened 2017-04-24
Add comment for flags field

vprithvi:


thrift/jaeger.thrift
```diff
@@ -58,7 +58,7 @@ struct Span {
   4:  required i64           parentSpanId # since nearly all spans will have parents spans, CHILD_OF refs do not have to be explicit
   5:  required string        operationName
   6:  optional list<SpanRef> references   # causal references to other spans
-  7:  required i32           flags        # tbd
+  7:  required i32           flags        # a bit field used to propagate sampling decisions. 1 signifies a SAMPLED span, 2 signifies a DEBUG span.
```
black-adder: 4 signifies deferred span?

vprithvi: It does, but that PR hasn't been merged yet, I'll add it in when it gets in!

yurishkuro: Rather than doing the comment, why don't we add the constants?

vprithvi: We can, or should we make these enums?

yurishkuro: the field has to remain `i32` to avoid a breaking change, so I don't think enums are useful.

vprithvi: I was thinking that the field remain `i32` but we assign enums to it. I was assuming that thrift enums are `i32` under the hood, but I'll verify.

PR #21, opened 2017-05-11
Add dependency thrift

oibe:


dependency.thrift
```diff
@@ -0,0 +1,45 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+namespace java com.uber.jaeger.thriftjava
```
black-adder: is it ok for namespaces to collide? pretty sure this namespace exists in another thrift file

dependency.thrift
```diff
@@ -0,0 +1,45 @@
...
+
+struct DependencyLink {
+  /** parent service name (caller) */
+  1: required string parent
+  /** child service name (callee) */
+  2: required string child
+  # 3: Moments OBSOLETE_duration_moments
```
black-adder: ?

dependency.thrift
```diff
@@ -0,0 +1,45 @@
...
+  /** calls made during the duration of this link */
+  4: required i64 callCount
+  # histogram?
+}
+
+/* An aggregate representation of services paired with every service they call. */
+struct Dependencies {
+  /** milliseconds from epoch */
+  1: required i64 start_ts
```
black-adder: why not camel case? what's the standard for thrift?

dependency.thrift
```diff
@@ -0,0 +1,45 @@
...
+
+struct DependencyLink {
+  /** parent service name (caller) */
+  1: required string parent
+  /** child service name (callee) */
+  2: required string child
+  # 3: Moments OBSOLETE_duration_moments
+  /** calls made during the duration of this link */
+  4: required i64 callCount
+  # histogram?
```
black-adder: don't feel like the thirft file is the best place for comments like this

dependency.thrift
oibe: It should be fine.  Its so that java programs now where to generate the name space.

thrift/dependency.thrift
```diff
@@ -0,0 +1,43 @@
...
+struct Dependencies {
+  // milliseconds from epoch
+  1: required i64 startTs
+  // milliseconds from epoch
+  2: required i64 endTs
+  3: required list<DependencyLink> links
+}
+
+service Dependency {
+    Dependencies getDependenciesForTrace(1: string traceId)
```
black-adder: shouldnt this function include startTs and endTs?

yurishkuro: why do we need this service? We are not going to expose it over thrift

thrift/dependency.thrift
```diff
@@ -0,0 +1,43 @@
...
+  // calls made during the duration of this link
+  4: required i64 callCount
+}
+
+// An aggregate representation of services paired with every service they call.
+struct Dependencies {
+  // milliseconds from epoch
+  1: required i64 startTs
+  // milliseconds from epoch
+  2: required i64 endTs
```
yurishkuro: what is the meaning of these timestamp fields? How are we going to use the?

thrift/dependency.thrift
yurishkuro: or do we plan to call this from Java over thrift/tchannel/yarpc?

thrift/dependency.thrift
oibe: I copied them from your original design, but it may be that we don't need them anymore since  this is more realtime.  Why were they in the original?

yurishkuro: The "original" wasn't my design, it was Zipkin's. In Zipkin this struct was used to represent aggregated dependencies for all services for a given time period (like 1 day). And even there it was largely redundant.

if you're adding some fields to the schema, you should have a clear use case for them.

thrift/dependency.thrift
oibe: @black-adder startTs and endTs were generated in my handler function, but are not needed anymore.
@yurishkuro I planned on calling this over thrift from the UDF.  Is there any reason not to?  Also I think this whole thing should be a "DataPipeline" service with "getDependenciesForTrace" as a single endpoint.

PR #22, opened 2017-05-13
Initial thrift for datapipeline service

oibe:


oibe:
@yurishkuro This is what I imagine the service setup will really be like.

thrift/datapipeline.thrift
```diff
@@ -0,0 +1,42 @@
...
+    // calls made during the duration of this link
+    4: required i64 callCount
+}
+
+// An aggregate representation of services paired with every service they call.
+struct Dependencies {
+    1: required list<DependencyLink> links
+}
+
+service Datapipeline {
```
yurishkuro: I don't think it is necessary to bundle unrelated endpoints in a single "service". We can easily have multiple services.  The `Dependencies` service will have a get endpoint and a save endpoint. Each service can have its own thrift file, with the relevant data structures.

PR #28, opened 2017-07-07
String is now required for the 2 files below.

oibe:


black-adder:
Probably should change the commit message

PR #30, opened 2017-07-12
Revert "Use map instead of list for baggage restrictions"

black-adder:
Reverts uber/jaeger-idl#29

1) thrift language support for maps is weird, prefer wire format to be as simple as possible. 
2) the list is more efficient than map
3) struct is more easy to extend if we wanted to return more fields in baggage restrictions, with map it's a breaking change

### DIFF
--- a/thrift/baggage.thrift
+++ b/thrift/baggage.thrift
@@ -20,9 +20,11 @@
 
 namespace java com.uber.jaeger.thriftjava
 
-typedef string BaggageKey
-# MaxValueLength is the maximum length of the baggage value.
-typedef i32 MaxValueLength
+# BaggageRestriction contains the baggage key and the maximum length of the baggage value.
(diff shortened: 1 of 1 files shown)


vprithvi:
@black-adder @yurishkuro Could you elaborate?

Specifically, why is the thrift support of maps weird? I went through the thrift bug [tracker](https://issues.apache.org/jira/browse/THRIFT-863?jql=project%20%3D%20THRIFT%20AND%20text%20~%20%22map%22), but most egregious map bugs seem to have been fixed. 

What is the efficiency gain in using thrift generated code to convert lists into maps as opposed to application logic; what do we do differently in our application that increases efficiency?

Maps support using structs too, one could easily have a Map<X, Y> where X, and Y are structs. One can freely add additional fields to the structs without breaking backward compatibility.

PR #31, opened 2017-07-24
Add thrift file health endpoint for datapipeline

oibe:


yurishkuro:
please elaborate why you need such detailed status over TChannel - who/what would be using it?

oibe:
@yurishkuro  yarpc uses the status defined here in order to determine its ok status.  For example the default status when starting up is "REFUSING", and when shutting down its "DRAINING".  In these states the healthcheck would respond with not ok.  Removing them seems like it would break the internal health logic involving yarpc.  However must servers define some sort of status like this, so perhaps we should adopt this.  If you want to remove this then how can we ensure that yarpc will not give invalid health checks?

yurishkuro:
we can do that internally, no need to expose this in jaeger-idl

also, see https://github.com/uber/jaeger/pull/280 with a different (http only) approach.

PR #33, opened 2017-10-10
Added netcore namespace

gideonkorir:
Added namespaces for netcore (C# with .netcore). Namespaces are:
1. ```Jaeger.Thrift```
2. ```Jaeger.Thrift.Agent```
3. ```Jaeger.Thrift.Agent.Zipkin```

I did not modify the build script because at the moment I'm using an [unofficial version of thrift that supports netcore](https://github.com/apache/thrift/pull/1379)

### DIFF
--- a/thrift/agent.thrift
+++ b/thrift/agent.thrift
@@ -24,6 +24,7 @@ include "jaeger.thrift"
 include "zipkincore.thrift"
 
 namespace java com.uber.jaeger.agent.thrift
+namespace netcore Jaeger.Thrift.Agent
 
 service Agent {
     oneway void emitZipkinBatch(1: list<zipkincore.Span> spans)
--- a/thrift/aggregation_validator.thrift
+++ b/thrift/aggregation_validator.thrift
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 namespace java com.uber.jaeger.thriftjava
+namespace netcore Jaeger.Thrift.Agent
 
 # ValidateTraceResponse returns ok when a trace has been written to redis.
 struct ValidateTraceResponse {
--- a/thrift/baggage.thrift
+++ b/thrift/baggage.thrift
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 namespace java com.uber.jaeger.thriftjava
+namespace netcore Jaeger.Thrift.Agent
 
 # BaggageRestriction contains the baggage key and the maximum length of the baggage value.
 struct BaggageRestriction {
--- a/thrift/crossdock/tracetest.thrift
+++ b/thrift/crossdock/tracetest.thrift
@@ -1,4 +1,5 @@
 namespace java com.uber.jaeger.crossdock.thrift
+namespace netcore Jaeger.Thrift.Crossdock
 
 enum Transport { HTTP, TCHANNEL, DUMMY }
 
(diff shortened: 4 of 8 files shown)


jeking3:
Does everything work as expected with the changes for dotnet 2.0 official from the thrift pull request?  I had some trouble getting the cross tests running a netcore server, but the netcore client is passing against all other thrift language servers that have common transports and protocols.

gideonkorir:
@jeking3 yes I managed to get the cross tests working alright (although I did have to ```pip install``` six and gevent to get the cross tests succeeding - the cpp tests were flaky every now and then). I saw the support in 1.0 but I wanted support for netstandard 2.0 because the new asp.net core libraries support netstandard 2.0 only and I was using them. In the future (assuming you merge your PR) I could help cross compile for netstandard 1.6 and 2.0.

jeking3:
Okay, I am going to merge this in, and any help you can provide expanding the usability to other .net core levels or in getting the servers running in cross test would be great.

thrift/agent.thrift
```diff
@@ -24,6 +24,7 @@ include "jaeger.thrift"
 include "zipkincore.thrift"
 
 namespace java com.uber.jaeger.agent.thrift
+namespace netcore Jaeger.Thrift.Agent
```
yurishkuro: since which Thrift version is this instruction supported?

jeking3: netcore is supported for dotnet-1.0.0-preview in Thrift 0.10.0
As of today, I merged in dotnet 2.0.0 official support into the thrift master which will become 0.11.0.

yurishkuro: Thanks. In this case we shouldn't merge this since most of our client libs are using thrift 0.9. 

We either need to strip this string during builds of other clients, or better have it added in dotnet project only.

gideonkorir: @yurishkuro I ran the thrift files against 0.9.2 and didn't get an issue. I did not alter the Makefile because of the version bump. At the moment I'm using the changes that just got merged to Master by @jeking3 so can't add a build file for that unless I use an unofficial thrift docker container for the build

jeking3:
The ubuntu-xenial Dockerfile in the thrift project includes almost all languages including dotnet-2.0 official and it is what we use to build most of the thrift jobs in Travis CI.

gideonkorir:
@jeking3 the build uses the ```thrift:0.92``` docker image. Unless I'm missing something the changes you merged into master aren't yet available as a docker image or am I wrong?

jeking3:
The docker image I described is the one we use to build thrift itself.  I'm not sure what's in the thrift:0.92 image or where it came from.

gideonkorir:
@jeking3 from what I can make from the Makefile we are using [this image](https://hub.docker.com/_/thrift/) to build the ```.thrift``` files. And that has up to version ```0.10.0``` at the time of this writing. As far as I can tell, the image has an already built thrift compiler are you suggesting we rebuild the compiler and then use it to run a Makefile for netcore? Thanks

jeking3:
That is not the official apache/thrift docker hub repository.  It looks like it is maintained by someone else.  The docker files that build it do not obtain files directly from the official github repo, but from a redistribution site or some kind.

isaachier:
Ya I don't know how that thrift old version (0.9.2) works for Go since there are breaking changes in the client library. I tried using a local checkout of 0.9.2 and got errors. @gideonkorir thanks for link to image.

PR #34, opened 2017-10-11
Add C++ namespaces.

isaachier:


thrift/agent.thrift
```diff
@@ -23,6 +23,7 @@
 include "jaeger.thrift"
 include "zipkincore.thrift"
 
+namespace cpp uber.jaeger.agent.thrift
```
yurishkuro: maybe `s/uber.jaeger/jaegertracing/`?

isaachier: Ya probably same applies to https://github.com/isaachier/cpp-client

thrift/sampling.thrift
```diff
@@ -20,6 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+namespace cpp jaegertracing.thrift.sampling_manager
```
yurishkuro: inconsistent with agent, it had thrift as a suffix

thrift/zipkincore.thrift
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+namespace cpp twitter.zipkin.thrift
```
yurishkuro: is this one necessary? I don't suppose you're planning to use zipkin model in C++ client.

thrift/sampling.thrift
isaachier: OK sure

thrift/zipkincore.thrift
isaachier: Honestly no. I can revert this. Just did this upfront and never went back to check what was used.

isaachier:
Actually, I need the Zipkin namespace in order to implement the Agent interface method `emitZipkinBatch` (even though all it does is throw an exception).

PR #40, opened 2017-11-16
PHP namespace.

damour:
For now we can set only global php namespace in generator:
  thrift -r --gen php:psr4,nsglobal=Jaeger\\ThriftGen ./

This is not very usable because some classes from jaeger.thrift missing as they has same name as zipkincore.thrift (e.g. Span)
Simple solution is to specify namespace for particular type.


damour:
@yurishkuro done.

PR #41, opened 2017-11-23
Add Protobuf support

isaachier:
Decided to spend some time on this. Not sure if we should do this now, but I'd love to swap out Thrift/UDP in the C++ client library in favor of Protobuf/gRPC.


protobuf/sampling.proto
```diff
@@ -0,0 +1,57 @@
...
+
+message ProbabilisticSamplingStrategy {
+    double sampling_rate = 1;
+}
+
+message RateLimitingSamplingStrategy {
+    int32 max_traces_per_seconds = 1;
+}
+
+message SamplingStrategy {
```
yurishkuro: I think this isn't a top level type, make sure you do proper nesting to minimize visibility

protobuf/sampling.proto
```diff
@@ -0,0 +1,57 @@
...
+message SamplingStrategyResponse {
+    oneof strategy {
+        ProbabilisticSamplingStrategy probabilistic = 1;
+        RateLimitingSamplingStrategy rate_limiting = 2;
+        PerOperationSamplingStrategy per_operation = 3;
+    }
+}
+
+service SamplingManager {
+    rpc GetSamplingStrategy (ServiceRequest) returns (SamplingStrategyResponse);
```
yurishkuro: with grpc it would make sense to use streaming for these comms

isaachier: So I was wondering in these types of cases where the request/response cycle is periodic but not exactly streaming. I guess it still works as long as we know we call it repeatedly.

protobuf/sampling.proto
isaachier: Good point.

protobuf/jaeger.proto
```diff
@@ -0,0 +1,81 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+package jaegertracing.protobuf;
+
+message Empty {};
+
+message ServiceRequest {
```
yurishkuro: can we keep this cleaner that thrift by separating data model from service definitions?

protobuf/jaeger.proto
```diff
@@ -0,0 +1,81 @@
...
+}
+
+message SpanRef {
+    enum Type {
+        CHILD_OF = 0;
+        FOLLOWS_FROM = 1;
+    }
+    Type type = 1;
+    uint64 trace_id_high = 2;
+    uint64 trace_id_low = 3;
```
yurishkuro: I think I would prefer these two fields to be encapsulated into a TradeId struct, however there is a question of efficiency in the generated Go types. For example, Thrift would define traceId field as a pointer in such case, which requires an additional allocation for no reason.

This is actually a more general question. Before committing the IDLs, we should really run perf tests to compare thrift/tchannel vs. proto/grpc, in terms of throughput, latency, memory, CPU.

protobuf/sampling.proto
```diff
@@ -0,0 +1,54 @@
...
+
+message ProbabilisticSamplingStrategy {
+    double sampling_rate = 1;
+}
+
+message RateLimitingSamplingStrategy {
+    int32 max_traces_per_seconds = 1;
+}
+
+message OperationSamplingStrategy {
```
yurishkuro: this can be embedded in PerOperationSamplingStrategy, correct?

protobuf/sampling.proto
```diff
@@ -0,0 +1,54 @@
...
+        ProbabilisticSamplingStrategy probabilistic = 2;
+        RateLimitingSamplingStrategy rate_limiting = 3;
+    }
+}
+
+message PerOperationSamplingStrategy {
+    double default_sampling_probability = 1;
+    double default_lower_bound_traces_per_second = 2;
+    repeated OperationSamplingStrategy per_operation_strategy = 3;
+    double default_upper_bound_traces_per_second = 4;
```
yurishkuro: nit: group with other primitive fields

protobuf/crossdock/tracetest.proto
```diff
@@ -0,0 +1,62 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+package jaegertracing.protobuf.crossdock;
+
+enum Transport {
```
yurishkuro: do you actually need this file right now? it's no longer an official Jaeger API, but a payload for crossdock tests, and those don't use proto/grpc. I would suggest leaving this out of this PR.

protobuf/sampling.proto
isaachier: Actually this is one of the more interesting issues with Protobuf v3. I think in Thrift, the `defaultUpperBoundTracesPerSecond` field is the only one marked as optional. Protobuf v3 makes all fields optional, so it makes sense. This default optional behavior may affect how we handle zero values.

protobuf/sampling.proto
isaachier: Guess so. I just copied this from Thrift. Then again, I suppose Thrift didn't support nesting.

protobuf/jaeger.proto
isaachier: Interesting question. For sure agree about the benchmarking. Actually started this change after reading about the different encoding/decoding benchmarks where it seems Protobuf has the best performance in Go, and presumably elsewhere. To set the record straight, when it comes to Go, there is the "fast" generator [gogo/protobuf](https://github.com/gogo/protobuf) vs. the standard generator [golang/protobuf](https://github.com/golang/protobuf). The reputation for speed is only true for the fast generator that I use throughout.

protobuf/jaeger.proto
isaachier: Sure. I would really call this `common.proto`. `oneway` isn't supported in gRPC (https://github.com/grpc/grpc/issues/2147), so need a way to return nothing. Regarding `ServiceRequest`, Protobuf does not support request or response types that are not defined as `message` types beforehand, so no `rpc Request(string service_name) returns (Response)`. Had to come up with this to wrap the string in two services (see `baggage.proto` and `sampling.proto`).

protobuf/crossdock/tracetest.proto
isaachier: No I don't. Just used it as an opportunity to port everything to Protobuf. Will remove for now.

isaachier:
Found out you can run `clang-format` on protobuf files so I did that now.

zkanda:
Hello @yurishkuro 

What do we need to move this PR forward?

I'd like to contribute on getting grpc integrated to Jaeger and this seems like one of the low hanging fruit. 

Cheers!

yurishkuro:
@zkanda @isaachier this PR is premature, please see https://github.com/jaegertracing/jaeger/issues/773

I think we should still keep this open since it contains useful build enhancements, and we can revisit it once some of the decisions are made in the above issue.

yurishkuro:
replaced by #61

PR #42, opened 2018-02-01
Fix permissions of generated files

isaachier:


black-adder:
Gonna need some context around this. So the generated thrift files now have user level permissions?

isaachier:
Yes. I run Linux on my desktop and `make clean` would tell me I didn't have permissions. Had to use `sudo make clean`. Swagger is already using this fix themselves, I just copied from that rule.

PR #61, opened 2020-03-11
Add jaeger proto files

pavolloffay:
Move the Jaeger proto files from the main repository here.


proto/model.proto
```diff
@@ -0,0 +1,167 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax="proto3";
+
+package jaeger.api_v2;
```
yurishkuro: should the `model.proto` be also in api_v2 folder? because other api_v2 files depend on it.

proto/model.proto
```diff
@@ -0,0 +1,167 @@
...
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+// TODO: document all types and fields
+
+// TODO: once this moves to jaeger-idl repo, we may want to change Go pkg to api_v2
+// and rewrite it to model only in this repo. That should make it easier to generate
+// classes in other languages.
```
yurishkuro: Any thoughts on how to address this? Specifically, how will this affect the backend, which has a large dependency footprint on /model/ package.

proto/model.proto
pavolloffay: It does not seem to be a problem for generating. 

I have able to generate query service in java https://github.com/jaegertracing/jaeger-analytics-java/blob/master/Makefile#L74

proto/model.proto
pavolloffay: How does it make easier to generate classes in other languages?

proto/model.proto
pavolloffay: I have moved it to `api_v2`, though I am not sure what implications it has

yurishkuro:
the key test for this PR is whether the main repo will work correctly by importing this revision as submodule.

pavolloffay:
I will test it out on Monday :)

pavolloffay:
I have pinned IDL to this branch `origin pull/61/head` in these two projects to verify the build:

* jaeger https://github.com/jaegertracing/jaeger/pull/2128
* jaeger-analytics-java https://github.com/jaegertracing/jaeger-analytics-java/pull/37

Makefile
```diff
@@ -43,4 +43,47 @@ thrift-image:
...
+		Mgoogle/api/annotations.proto=github.com/gogo/googleapis/google/api, \
+		Mmodel.proto=github.com/jaegertracing/jaeger/model \
+	| sed 's/ //g')
+
+.PHONY: proto
+proto:
+	mkdir -p proto-gen
+	$(PROTOC) \
+		$(PROTO_INCLUDES) \
+		--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/proto-gen \
```
yurishkuro: if think we should be trying to build for multiple languages (see #41)

pavolloffay: +1 in fact I didn't know it can be generated into multiple lags using a single command

pavolloffay: I have made the directories to varios langs configurable. The makefile could be loaded in jaeger and just changed the output dirs

PR #63, opened 2020-04-15
Improve csharp support

### LINKED ISSUE #62: Improve C# support
## Requirement - what kind of business use case are you trying to solve?

We need to get `proto-gen-csharp` usable. As of now, `proto-gen-csharp` can not be used because `google/api/annotations.proto`, `gogoproto/gogo.proto` and `protoc-gen-swagger/options/annotations.proto` are not created but necessary for reflection (which can't be removed).

## Problem - what in Jaeger blocks you from solving the requirement?

`gogoproto/goto.proto` can't be generated in C# since it's using proto2 syntax which only got added with Protocol Buffers v3.11.0 which was available starting gRPC v1.27.0. This docker currently targets 1.26.0.

This is blocked by https://github.com/jaegertracing/docker-protobuf/issues/14.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Update the `Makefile` to generate `google/api/annotations.proto`, `gogoproto/gogo.proto` and `protoc-gen-swagger/options/annotations.proto`.

## Any open questions to address

I will create the pull request myself.

Falco20019:
- Fixes #62 
- Blocked by https://github.com/jaegertracing/docker-protobuf/pull/15

* Thrift doesn't support netcore in 0.13.0 anymore
* C# can't be generated with 0.9.2, but 0.13.0 can't be used as blocked by https://github.com/ahawkins/docker-thrift/pull/22
  * Therefore only preparations made
* Removed unused import in model.proto to avoid warnings
* Improved Makefile to make it easier to read and maintain
* Generate annotation protos necessary in C# for reflection

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ PROTO_GEN_CSHARP_DIR ?= proto-gen-csharp
 PROTOC_WITHOUT_GRPC := $(PROTOC) \
 		$(PROTO_INCLUDES) \
 		--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/${PROTO_GEN_GO_DIR} \
-		--java_out=proto-gen-java \
+		--java_out=${PROTO_GEN_JAVA_DIR} \
 		--python_out=${PROTO_GEN_PYTHON_DIR} \
 		--js_out=${PROTO_GEN_JS_DIR} \
 		--cpp_out=${PROTO_GEN_CPP_DIR} \
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ protocompile:
 
 PROTO_INCLUDES := \
 	-Iproto/api_v2 \
+	-Iproto \
 	-I/usr/include/github.com/gogo/protobuf
 # Remapping of std types to gogo types (must not contain spaces)
 PROTO_GOGO_MAPPINGS := $(shell echo \
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,15 @@ THRIFT=docker run -u $(shell id -u) -v "${PWD}:/data" $(THRIFT_IMG) thrift
 
 SWAGGER_VER=0.12.0
 SWAGGER_IMAGE=quay.io/goswagger/swagger:$(SWAGGER_VER)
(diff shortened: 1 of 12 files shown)


thrift/agent.thrift
```diff
@@ -18,7 +18,7 @@ include "zipkincore.thrift"
 namespace cpp jaegertracing.agent.thrift
 namespace java io.jaegertracing.agent.thrift
 namespace php Jaeger.Thrift.Agent
-namespace netcore Jaeger.Thrift.Agent
+namespace netstd Jaeger.Thrift.Agent
```
yurishkuro, line 21: @Falco20019 could you elaborate on this change? Is it possible that both namespaces can remain?

Makefile
```diff
@@ -65,78 +68,51 @@ PROTO_GEN_JS_DIR ?= proto-gen-js
...
-		${PROTO_GEN_PYTHON_DIR} \
-		${PROTO_GEN_JS_DIR} \
-		${PROTO_GEN_CPP_DIR} \
-		${PROTO_GEN_CSHARP_DIR}
-
-	$(PROTOC) \
+PROTOC_WITHOUT_GRPC := $(PROTOC) \
 		$(PROTO_INCLUDES) \
 		--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/${PROTO_GEN_GO_DIR} \
 		--java_out=proto-gen-java \
```
yurishkuro, line 74: let's use `${PROTO_GEN_JAVA_DIR}`

thrift/agent.thrift
Falco20019: `netcore` is not available in Thrift 0.13 anymore. It was completely replaced with `netstd` which is compatible with .NET Core and .NET Framework (to some extend).

Makefile
Falco20019: Thanks, I replaced it :)

thrift/agent.thrift
Falco20019: I added both. In my tests I saw, that depending on 0.9.2 or 0.13.0 you will get the warning for either of them. 0.9.2 does not know `netstd`...

Falco20019:
#64 is in, so once jaegertracing/docker-protobuf#15 is released, this becomes testable.

PR #64, opened 2020-06-25
Use Thrift 0.13.0

Falco20019:
- https://github.com/jaegertracing/jaeger/issues/1972
- https://github.com/uber/tchannel-go/issues/781
- https://github.com/jaegertracing/jaeger-client-csharp/issues/139
- https://github.com/jaegertracing/jaeger-idl/issues/62

## Which PRs is this blocked on?
- https://github.com/jaegertracing/jaeger-idl/pull/63
- https://github.com/jaegertracing/docker-protobuf/pull/15

- The Go and C# clients need to upgrade Thrift to `0.13`. Go because of a inherited Prometheus dependency, C# because of strong naming assemblies. The current `jaeger-idl` only produces `0.9.2`. Right now, we have to manually adjust the `Makefile` to use the necessary version. There is no way to also use this as submodule.

I would like to check if other libraries need to stay on `0.9.2` for any reason or if we can upgrade this for all. If any have to stay on the old version, we might modify the Makefile to use different versions for the different clients.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 
-THRIFT_VER=0.9.2
-THRIFT_IMG=thrift:$(THRIFT_VER)
+THRIFT_VER=0.13
+THRIFT_IMG=jaegertracing/thrift:$(THRIFT_VER)
 THRIFT=docker run -u $(shell id -u) -v "${PWD}:/data" $(THRIFT_IMG) thrift
 
 SWAGGER_VER=0.12.0
@@ -28,7 +28,7 @@ swagger-validate:
 	$(SWAGGER) validate ./swagger/zipkin2-api.yaml
 
 clean:
-	rm -rf gen-* || true
+	rm -rf *gen-* || true
 
 thrift:	thrift-image clean $(THRIFT_FILES)
 
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 
-THRIFT_VER=0.13
-THRIFT_IMG=jaegertracing/thrift:$(THRIFT_VER)
+THRIFT_VER?=0.13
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 
-THRIFT_VER=0.9.2
-THRIFT_IMG=thrift:$(THRIFT_VER)
+THRIFT_VER=0.13
+THRIFT_IMG=jaegertracing/thrift:$(THRIFT_VER)
```
pavolloffay, line 3: These two should be made configurable to allow submodules modify the build.

```
THRIFT_VER ?= 
```

pavolloffay:
The client libraries can use an arbitrary thrift version. The version set in this repository is used for the CI job. 

We should make the version configurable so anybody pulling the makefile from this repository can override it.

Makefile
Falco20019: Currently only the directories are configurable. I currently reworked a lot of the `Makefile` in https://github.com/jaegertracing/jaeger-idl/pull/63 but this still doesn't make it configurable.

We might want to have a separate PR for making all docker versions configurable. I tried to keep the change in this PR minimal.

pavolloffay: >  I tried to keep the change in this PR minimal.

The changes I am asking for are for the lines that are changed in this PR.

I am not asking to change other docker versions just this one. The issue is that the `jaegertracing/thrift` contains only a single version.

Falco20019: I'm fine with it. Adjusted it accordingly :) Might still be a good idea to discuss having all the others also configurable in a separate PR once this is in.

pavolloffay:
There is a conflict in the Makefile

pavolloffay:
thanks @Falco20019

Falco20019:
Thank you! :)

PR #73, opened 2021-03-31
Add metric def and metrics query endpoints

albertteoh:
Signed-off-by: albertteoh <albert.teoh@logz.io>

- The first of multiple PRs addressing: https://github.com/jaegertracing/jaeger/issues/2736

- The API specification for the jaeger-query `/metrics` endpoint, containing sub paths leading to specific R.E.D metrics aggregations.
- Design goals:
  - Align the API to the requirements for the use case which are to support retrieval of R.E.D metrics aggregated on a per-service level and on a per-operation level (when a service_name is selected).
  - Aim for a simplicity and intuitiveness.
- Sample HTTP requests; for each of these examples, optional parameters can be added such as `endTs=<posix millis>&lookback=<duration millis>&step=<duration millis>&rate=<duration millis>`:
  - `curl -i http://localhost:16686/api/metrics/latencies?quantile=0.95`
  - `curl -i http://localhost:16686/api/metrics/calls`
  - `curl -i http://localhost:16686/api/metrics/errors`
  - `curl -i http://localhost:16686/api/metrics/latencies/myservice?quantile=0.95`
  - `curl -i http://localhost:16686/api/metrics/calls/myservice`
  - `curl -i http://localhost:16686/api/metrics/errors/myservice`

- Import [OTEL's metrics wire format](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto), with an addition of `BucketUpperBound`; motivation added as comment in protobuf.
  - **Rationale**: Considered using [OpenMetrics' protobuf wire format](https://github.com/OpenObservability/OpenMetrics/blob/main/proto/openmetrics_data_model.proto). However, given [OpenTelemetry's goal of supporting both Prometheus and StatsD](https://medium.com/opentelemetry/opentelemetry-metrics-roadmap-f4276fd070cf) the hope is that adopting OpenTelemetry's data model will enable us to more easily support non-Prometheus compatible metrics backends in future with less effort and lower likelihood of breaking changes; albeit at the cost of less stability in the short-term. Having said that, OpenTelemetry's spec is more than adequate for our relatively simple requirements of using just 2 well-supported metrics types: Counters and Histograms. As such, I don't foresee a dire need to maintain parity with OpenTelemetry over the coming months as it will no doubt continue evolving towards a more stable API and data model.
  - Taking the import instead of copy approach of referencing otel's metrics proto to reduce maintenance overhead and chances for diverging versions.


proto/api_v2/model.proto
```diff
@@ -171,3 +171,648 @@ message DependencyLink {
...
+  // in this bucket are less than or equal to (<=) the BucketUpperBound value.
+  //
+  // Motivation: We want to offer an HTTP/JSON endpoint for querying metrics.
+  // An "+Inf" bucket is implicitly added by prometheus client to histograms and,
+  // as JSON libraries cannot marshal +Inf bucket values, adding BucketUpperBound enables
+  // implementing the MarshalJSON function to stringify the "+Inf" bucket that is returned
+  // back from promQL queries.
+  message BucketUpperBound {
+    double upper_bound = 1;
+  }
```
albertteoh, line 675: This was the only meaningful modification (addition) to the [OpenTelemetry metrics wire format](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto). 

Let me know if there is a better alternative to solving the JSON marshalling of `+Inf` values, or if I'm going down the wrong path somehow.

yurishkuro (commented):
A few general comments:

- I don't think it's necessary to mix metrics stuff with the primary model and QueryService, we can do them as separate files.

- Placing data model in this repo signals long term support, I think it's premature. We can host the files directly in Jaeger repo, as we used to do with api_v2. Especially because afaik the OTEL metrics model is not fixed yet.

- Could we import OTEL model somehow instead of copying it?

- Open-ended prom query string makes me anxious, as it exposes a very wide api. Can we not do something more specific and narrow?

albertteoh:
Thanks for the quick feedback, @yurishkuro.

> I don't think it's necessary to mix metrics stuff with the primary model and QueryService, we can do them as separate files.

To confirm, you'd also like the metrics rpc methods in a separate service outside of QueryService (i.e. a new "MetricsQueryService")? 

> Placing data model in this repo signals long term support, I think it's premature. We can host the files directly in Jaeger repo, as we used to do with api_v2. Especially because afaik the OTEL metrics model is not fixed yet.

Okay, I can move the service/rpc methods into the Jaeger repo. Would [model/proto](https://github.com/jaegertracing/jaeger/tree/master/model/proto) be a good home for these?

Yes, that's correct, OTEL metrics spec is in the works with a [stable version around September 2021](https://medium.com/opentelemetry/opentelemetry-metrics-roadmap-f4276fd070cf), if I'm not mistaken.

> Could we import OTEL model somehow instead of copying it?

Yes, I think we can import it. Is the best approach to add otel metrics proto in https://github.com/jaegertracing/docker-protobuf, which `make proto` (from jaegertracing/jaeger) references for generating Go code?

fwiw, the main reason for copying was to workaround the `Inf` marshalling to JSON, but I think there's a better way to do this without needing to modify the OTEL metrics data model.

> Open-ended prom query string makes me anxious, as it exposes a very wide api. Can we not do something more specific and narrow?

One thing I could do is remove the `custom` type, which is a true open-ended prom query. 

However, with `calls`, `errors` and `latency`, we'd want to give the caller the ability to search for one or more timeseries within the restricted set of known R.E.D. metric names (specifically, `calls_total`, `latency_count`, `latency_sum` and `latency_bucket`), which could contain any number of labels. Callers will not be able to query any other metric name.

Another possible restriction is to enforce strict equality matching of labels instead of allowing any promQL label filter expression within the [range of available operators](https://prometheus.io/docs/prometheus/latest/querying/basics/).

So instead of allowing something like `calls_total{service_name=~".*foo.*"}`, callers would only be able to do `calls_total{service_name="foo"}`

yurishkuro:
> To confirm, you'd also like the metrics rpc methods in a separate service outside of QueryService (i.e. a new "MetricsQueryService")?

yes.

proto/api_v2/query.proto
```diff
@@ -124,6 +124,62 @@ message GetDependenciesResponse {
...
+message GetMetricsResponse {
+  repeated jaeger.api_v2.Metric metrics = 1 [
+    (gogoproto.nullable) = false
+  ];
+}
+
+// The type of aggregated metric.
+enum AggregatedMetricType {
+  // The number of calls/requests made.
+  CALLS = 0;
```
yurishkuro, line 173: iirc Protobuf (or Google) recommends reserving 0 for "undefined"

yurishkuro:
> Okay, I can move the service/rpc methods into the Jaeger repo. Would model/proto be a good home for these?

yeah, it's fine. When it's internal in the repo it's much easier to move around if we need to.

> Yes, I think we can import it. Is the best approach to add otel metrics proto in https://github.com/jaegertracing/docker-protobuf, which make proto (from jaegertracing/jaeger) references for generating Go code?

Another option is to create an `otel-idl` submodule, similar to `idl` that links to `jaeger-idl`, and arrange the build to import from there. Submodule still gives us versioning against future changes in OTEL. Adding to docker-protobuf is a good idea (as submodules are easy to add, hard to remove), it's just a bit more friction on getting upgrades.

Re the query, I would suggest trying to formulate the exact requirements and define an IDL just for that, without resorting to free-style query strings like `calls_total{service_name="foo"}`. I think this service has very specific requirements about which time series it should support.

About using the enum - please check that enums are extensible in Protobuf. In Thrift they are not extensible without sync-upgrade of client and server.

albertteoh:
> Adding to docker-protobuf is a good idea (as submodules are easy to add, hard to remove), it's just a bit more friction on getting upgrades.

I don't imagine we would need to upgrade frequently, if at all. The existing data model offered by OpenTelemetry satisfies all our current requirements, unless of course we wish to introduce new requirements that depend on a new data model.
 
> Re the query, I would suggest trying to formulate the exact requirements and define an IDL just for that, without resorting to free-style query strings like `calls_total{service_name="foo"}`. I think this service has very specific requirements about which time series it should support.

There are specific requirements relating to the _metric name_ restricted to just "calls" (in which "errors" belong) and "latencies". 
However there are no requirements/constraints surrounding supported time series within these two metric groups. For example, one time series could be `calls_total{service_name=foo}`, another `calls_total{service_name=bar}` and another `calls_total{service_name=foo, operation=/foo}`, and all three of these are unique and separate time series. The caller's query should be able to fetch any of these, returning 0 to many time series along with their data points.

However, I agree that we should remove the promQL query syntax, and I propose that we adopt the same API as jaeger-query currently exposes for [FindTraces tags filter](https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/query.proto#L65) that uses a `map<string, string>` (e.g. `{"service_name": "foo",  "operation": "/foo"}`) which achieves the following goals:
- Still satisfies our functional requirement of allowing users to filter by metric dimensions/labels.
- Constrains API.
- Makes the API explicit and easier to infer what values to pass rather than just a vague "string".
- Consistent experience with an existing jaeger-query API, which is tried and tested.
- Agnostic to a particular tech/query language (prometheus/promQL) and defer the construction of the actual query to the implementing reader layer.

> About using the enum - please check that enums are extensible in Protobuf. In Thrift they are not extensible without sync-upgrade of client and server.

I tested this out and it appears to be extensible. Either client or server side can be upgraded (new fields added to an enum). If the new field/s are used by either side while the other hasn't been upgraded, the enum int would still be received without error, but not available in code and so we'll need to handle this gracefully.

proto/api_v2/query.proto
albertteoh: You're right, good catch.

yurishkuro:
> For example, one time series could be calls_total{service_name=foo}, another calls_total{service_name=bar} and another calls_total{service_name=foo, operation=/foo}, and all three of these are unique and separate time series.

What I would suggest is have a logical description of these metrics in the service API, similar to that enum (CALLS vs. ERRORS vs. LATENCY, etc.) Then in the prom-based implementation of the API map them to specific prom metrics as needed. You can even have the implementation take a configuration file that contains the mapping from logical API terms into exact metric queries.

albertteoh:
> What I would suggest is have a logical description of these metrics in the service API, similar to that enum (CALLS vs. ERRORS vs. LATENCY, etc.) Then in the prom-based implementation of the API map them to specific prom metrics as needed. You can even have the implementation take a configuration file that contains the mapping from logical API terms into exact metric queries.

Yes, if I understand you correctly, this is precisely the plan with the enum types. Some examples of mappings from logical API to promQL queries ([documented here](https://docs.google.com/document/d/1JVW-Mjn4QCqYJKeyCVIWNbXRdeGbqJjghKuIcIBqHKo/edit#heading=h.narc1ywxhh7z)):

**CALLS**
http://localhost:16686/api/metrics?type='calls'&labels='service_name="currencyservice",operation="/GetConversion"'
  -> `calls_total{service_name="currencyservice", operation="/GetConversion"}`

**ERRORS**
http://localhost:16686/api/metrics?type='errors'&labels='service_name="currencyservice",operation="/GetConversion"' 
  -> `calls_total{service_name="currencyservice", operation="/GetConversion", status_code="ERROR"}`

**LATENCY**
http://localhost:16686/api/metrics?type='latency'&labels='service_name="currencyservice",operation="/GetConversion"'
  -> `latency_buckets{service_name="currencyservice", operation="/GetConversion"}`
  -> `latency_count{service_name="currencyservice", operation="/GetConversion"}`
  -> `latency_sum{service_name="currencyservice", operation="/GetConversion"}`

Some additional comments:
- The key-value `labels` are synonymous with how we do tag-based searches in Jaeger.
- It's not a simple 1:1 mapping between enum and metrics and so I'm not sure about making it configurable. The goal is to make the API as logically simple as possible and defer the more complex nuances of fetching the correct metrics to the underlying implementation.

albertteoh:
Please ignore the last couple of pushes, they were accidental.

albertteoh:
@yurishkuro I've removed the the copied otel metrics wire format and imported it from the newly created `MetricsQueryService` as discussed.

As per the recommendation in https://github.com/open-telemetry/opentelemetry-proto:
> The proto files can be consumed as GIT submodule or copied over and built directly in the consumer project.

Since we prefer to import rather than copy the data model, I've tested the approach of creating a submodule referencing opentelemetry-proto and managed to get it working; albeit with a workaround required for the import references, which I'll show and discuss in a subsequent PR.

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,94 @@
...
+import "model.proto";
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+import "opentelemetry/proto/metrics/v1/metrics.proto";
+
+option go_package = "api_v2";
+option java_package = "io.jaegertracing.api_v2";
+
```
albertteoh, line 28: I could not include the `marshalto`, etc. [Gogo proto extensions](https://github.com/gogo/protobuf/blob/master/extensions.md) as [query.proto does](https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/query.proto#L29-L35) because the OTEL metrics doesn't adopt them, leading to incompatibilities.

albertteoh:
Setting this PR to draft as I'll need to think about how users can take advantage of promQL's functions such as quantile calculations and rates, which I think would be key requirements to a potential future Jaeger UI.

albertteoh:
> What I would suggest is have a logical description of these metrics in the service API, similar to that enum (CALLS vs. ERRORS vs. LATENCY, etc.)

@yurishkuro I've applied your suggestion to align the API more closely to the requirements as part of the aggregated views on R.E.D. metrics. 

The API now provides "pre-packaged" promQL queries for specific use cases via dedicated rpc functions/URL paths, instead of placing the onus on the caller to build the queries themselves; so it looks more explicit (while being configurable through optional parameters) and intuitive, and has the added benefit of decoupling it from prometheus; thank you for the advice!

Please let me know what you think and if it could be improved further.

albertteoh:
> Placing data model in this repo signals long term support, I think it's premature. We can host the files directly in Jaeger repo, as we used to do with api_v2. Especially because afaik the OTEL metrics model is not fixed yet.

I may have misunderstood you here, thinking that if we were to copy the OTEL metrics model, that you'd prefer the copied data model in the Jaeger repo, and leave the MetricsQueryService here in the jaeger-idl repo given this is the home of the existing jaeger-query API definitions.

However, if you prefer the changes in this PR to be in the Jaeger repo as well, please let me know and I'll look into moving it there.

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,158 @@
...
+  google.protobuf.Duration lookback = 2 [
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = true
+  ];
+
+  // step size is the duration between data points of the query results.
+  // For example, if set to 5s, the results would produce a data point every 5 seconds
+  // from the start_time to end_time.
+  // Optional. Default = 5s.
+  google.protobuf.Duration step = 3 [
```
yurishkuro, line 50: Why do you need this? Why not just accept whichever precision the backend can return?

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,158 @@
...
+  // service_name is the service name to fetch operation metrics from.
+  // Required.
+  string service_name = 2;
+}
+
+service MetricsQueryService {
+  // GetServiceLatencies gets the per-service latencies.
+  rpc GetServiceLatencies(GetServiceLatenciesRequest) returns (GetMetricsResponse) {
+    option (google.api.http) = {
+      get: "/metrics/latencies"
```
yurishkuro, line 120: I wouldn't bother with http, we're not using it anywhere.

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,158 @@
...
+  // from the start_time to end_time.
+  // Optional. Default = 5s.
+  google.protobuf.Duration step = 3 [
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = true
+  ];
+
+  // rate is the duration in which the per-second rate of change is calculated for a cumulative counter metric.
+  // Optional. Default = 10m.
+  google.protobuf.Duration rate = 4 [
```
yurishkuro, line 57: Confusing name

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,158 @@
...
+// GetServiceLatenciesRequest contains parameters for the GetServiceLatencies RPC call.
+message GetServiceLatenciesRequest {
+  MetricsQueryBaseRequest baseRequest = 1;
+  // quantile is the quantile to compute from latency histogram metrics.
+  // Required.
+  double quantile = 2;
+}
+
+// GetServiceCallRatesRequest contains parameters for the GetServiceCallRates RPC call.
+message GetServiceCallRatesRequest {
```
yurishkuro, line 80: Which service?

proto/api_v2/metricsquery.proto
```diff
@@ -0,0 +1,158 @@
...
+  MetricsQueryBaseRequest baseRequest = 1;
+}
+
+// GetServiceErrorRatesRequest contains parameters for the GetServiceErrorRates RPC call.
+message GetServiceErrorRatesRequest {
+  MetricsQueryBaseRequest baseRequest = 1;
+}
+
+// GetOperationLatenciesRequest contains parameters for the GetOperationLatencies RPC call.
+message GetOperationLatenciesRequest {
```
yurishkuro, line 90: Which operation?

yurishkuro (commented):
Yes I recommend prototyping directly in the main repo.

proto/api_v2/metricsquery.proto
albertteoh: Required by [M3's query API](https://m3db.io/docs/reference/m3query/api/query/):
```
curl -X "POST" -G "http://localhost:7201/api/v1/query_range" \
   -d "query=calls_total" \
   -d "start=$( date -v -45S +%s )" \
   -d "end=$( date +%s )" | jq .
   
{
  "status": "error",
  "error": "error parsing param: step, error: expected positive step size, instead got: 0"
}
```

Further, I think it's useful for callers to control the amount of data downloaded from the query engine. 

What if metrics data is written at a higher precision in the timeseries db backend, or the timeseries backend is able to interpolate data points at a higher default precision, than the caller desires because it doesn't need such a high precision at the cost of performance?

albertteoh (commented):
Thank for the feedback, I'll close this PR and move the changes over to the main jaeger repo once I've addressed your comments.

proto/api_v2/metricsquery.proto
yurishkuro: Ok. But I'm concerned with the reverse, client asking for precision that server doesn't have, what is the expected behavior? This param can only be satisfied if it's equal to or a multiple of the server's native precision, and how is client supposed to know that?

proto/api_v2/metricsquery.proto
yurishkuro: ratePer: 5min

proto/api_v2/metricsquery.proto
yurishkuro: How do you envision using this API? Eg in an architecture with thousands of services?  It seems you want to pass service names as args, not expect to receive all services.

proto/api_v2/metricsquery.proto
yurishkuro: Well, you may need to try this out then. I tried that using grpc-gateway lib, I don't know if it's still the recommended way, and it was certainly weird back then, requiring you to run http server that would proxy the request to grpc server, even when they were in the same process.

proto/api_v2/metricsquery.proto
albertteoh: Firstly, I'll take a step back and clarify some sample use cases I have in mind that apply to this API:
- An org-level error rate SLA alert is triggered. How can we root cause this?
  - What services have higher error rates? Are their QPS's significant enough for us to care about?
  - Once we've identified the candidate services, we can drill down into their operations, then further to their traces to pinpoint the child services/operations that are raising the errors and mitigate accordingly.

- Capacity planning:
  - What services have the highest QPSs and latencies? These could be candidate services to target for measuring the bulk of resource consumers.
  - Drilling down into their traces, we can identify the downstream services that incur the bulk of the time and data stores they use.
  - We could view long-term patterns of QPSs and latencies for these downstream services/data stores and anticipate how much more capacity is required for these resource-intensive services.
  - At the same time, we may even identify optimizations on these resource hungry services.

In both cases, we can't easily know, upfront, the service/s with the highest latencies, error or request rates; instead, relying on a sorted view of aggregated metrics grouped by service, which is what I envision this API to provide.

proto/api_v2/metricsquery.proto
albertteoh: My mistake, I was confusing our native HTTP endpoints (port 16686) with the gRPC (port 16685) HTTP endpoints. Makes sense to not bother with HTTP on the gRPC listener; I'll remove it.

proto/api_v2/metricsquery.proto
yurishkuro: The client of the API doesn't know about the backend implementation. The only possible approach I see is if there is another api call that returns the supported precisions and the ui makes the user choose

albertteoh (commented):
Thanks again for the input, @yurishkuro. I'll close this PR and move this to the main jaeger repo.

PR #76, opened 2021-06-10
Add gRPC query service with OTLP model

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

Related to https://github.com/jaegertracing/jaeger/issues/169#issuecomment-856625509

### DIFF
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -29,7 +29,7 @@ message GetTraceRequest {
   string trace_id = 1;
 }
 
-// A single response chunk holds spans from a single trace.
+// Response object with spans.
 message SpansResponseChunk {
   // A list of OpenTelemetry ResourceSpans.
   // In case of JSON format the ids (trace_id, span_id, parent_id) are encoded in base64 even though OpenTelemetry specification
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -48,11 +48,13 @@ message TraceQueryParameters {
   // Attributes are matched against Span and Resource attributes.
   // At least one span in a trace must match all specified attributes.
   map<string, string> attributes = 3;
+  // Span min start time.
   google.protobuf.Timestamp start_time_min = 4;
+  // Span max start time.
   google.protobuf.Timestamp start_time_max = 5;
-  // Minimal duration of a span.
+  // Span max duration.
   google.protobuf.Duration duration_min = 6;
-  // Maximal duration of a span.
+  // Span min duration.
   google.protobuf.Duration duration_max = 7;
   // Maximum number of traces in the response.
   // Note that some storages (e.g. Cassandra) does not guarantee this parameter.
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -29,6 +29,8 @@ message GetTraceRequest {
 }
 
 // A single response chunk holds a single trace.
+// Note that a subsequent request to get a trace might return a different result
+// because spans are reported in asynchronous chunks.
 message SpansResponseChunk {
   // OTLP resource spans
   // see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -29,7 +29,7 @@ message GetTraceRequest {
   string trace_id = 1;
 }
 
-// A single response chunk holds a single trace.
+// A single response chunk holds spans from a single trace.
 // Note that a subsequent request to get a trace might return a different result
 // because spans are reported in asynchronous chunks.
 message SpansResponseChunk {
@@ -45,11 +45,17 @@ message SpansResponseChunk {
 message TraceQueryParameters {
   string service_name = 1;
   string operation_name = 2;
+  // Attributes are matched against Span and Resource attributes.
+  // At least one span in a trace must match all specified attributes.
   map<string, string> attributes = 3;
   google.protobuf.Timestamp start_time_min = 4;
   google.protobuf.Timestamp start_time_max = 5;
+  // Minimal duration of a span.
   google.protobuf.Duration duration_min = 6;
+  // Maximal duration of a span.
   google.protobuf.Duration duration_max = 7;
+  // Maximum number of traces in the response.
+  // Note that some storages (e.g. Cassandra) does not guarantee this parameter.
   int32 num_traces = 8;
 }
 
@@ -68,7 +74,9 @@ message GetServicesResponse {
 
 // Request object to search for operation names.
 message GetOperationsRequest {
+  // Required service name.
   string service = 1;
+  // Optional span kind.
   string span_kind = 2;
 }
 
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -25,7 +25,7 @@ option go_package = "api_v3";
 option java_package = "io.jaegertracing.api_v3";
 
 message GetTraceRequest {
-  bytes trace_id = 1;
+  string trace_id = 1;
 }
 
 message SpansResponseChunk {
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -31,7 +31,7 @@ message GetTraceRequest {
 
 // A single response chunk holds spans from a single trace.
 message SpansResponseChunk {
-  // OTLP resource spans [1].
+  // A list of OpenTelemetry ResourceSpans.
   // In case of JSON format the ids (trace_id, span_id, parent_id) are encoded in base64 even though OpenTelemetry specification
   // mandates to use hex encoding [2].
   // Base64 is chosen to keep compatibility with JSONPb codec.
@@ -73,7 +73,7 @@ message GetServicesResponse {
   repeated string services = 1;
 }
 
-// Request object to search for operation names.
+// Request object to get operation names.
 message GetOperationsRequest {
   // Required service name.
   string service = 1;
@@ -93,17 +93,15 @@ message GetOperationsResponse {
 }
 
 service QueryService {
-  // GetTraces returns single trace.
+  // GetTrace returns a single trace.
   // Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
-  // and so the response cannot be directly unmarshalled. This can be fixed by standard JSON
-  // parsing libraries or string manipulation to remove the envelope.
+  // It means that the JSON response cannot be directly unmarshalled using JSONPb.
+  // This can be fixed by first parsing into user-defined envelope with standard JSON library
+  // or string manipulation to remove the envelope. Alternatively generate objects using OpenAPI.
   rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
 
   // FindTraces searches for traces.
-  // Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
-  // Hence the response cannot be directly unmarshalled into response protobuf type.
-  // This can be fixed by first unmarshalling into user defined envelope type using standard JSON library
-  // or by removing the envelope by string manipulation.
+  // See GetTrace for JSON unmarshalling.
   rpc FindTraces(FindTracesRequest) returns (stream SpansResponseChunk) {}
 
   // GetServices returns service names.
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -1,0 +1,78 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax="proto3";
+
+package jaeger.api_v3;
+
+import "opentelemetry/proto/trace/v1/trace.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "api_v3";
+option java_package = "io.jaegertracing.api_v3";
+
+message GetTraceRequest {
+  bytes trace_id = 1;
+}
+
+message SpansResponseChunk {
+  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
+}
+
+message TraceQueryParameters {
+  string service_name = 1;
+  string operation_name = 2;
+  map<string, string> tags = 3;
+  google.protobuf.Timestamp start_time_min = 4;
+  google.protobuf.Timestamp start_time_max = 5;
+  google.protobuf.Duration duration_min = 6;
+  google.protobuf.Duration duration_max = 7;
+  int32 search_depth = 8;
+}
+
+message FindTracesRequest {
+  TraceQueryParameters query = 1;
+}
+
+message GetServicesRequest {}
+
+message GetServicesResponse {
+  repeated string services = 1;
+}
+
+message GetOperationsRequest {
+  string service = 1;
+  string span_kind = 2;
+}
+
+message Operation {
+  string name = 1;
+  string span_kind = 2;
+}
+
+message GetOperationsResponse {
+  repeated Operation operations = 2;
+}
+
+service QueryService {
+  rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
+
+  rpc GetTraces(FindTracesRequest) returns (stream SpansResponseChunk) {}
+
+  rpc GetServices(GetServicesRequest) returns (GetServicesResponse) {}
+
+  rpc GetOperations(GetOperationsRequest) returns (GetOperationsResponse) {}
+}
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -41,7 +41,7 @@ message SpansResponseChunk {
 }
 
 // Query parameters to find traces.
-// Note that some storage implementations do not guarantee correct implementation of all parameters.
+// Note that some storage implementations do not guarantee the correct implementation of all parameters.
 message TraceQueryParameters {
   string service_name = 1;
   string operation_name = 2;
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -25,6 +25,7 @@ option java_package = "io.jaegertracing.api_v3";
 
 // Request object to get a trace.
 message GetTraceRequest {
+  // Hex encoded 128 bit trace_id
   string trace_id = 1;
 }
 
@@ -34,6 +35,9 @@ message GetTraceRequest {
 message SpansResponseChunk {
   // OTLP resource spans
   // see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto
+  // In case of JSON format the ids (trace_id, span_id, parent_id) are encoded in base64 even though OpenTelemetry specification
+  // mandates to use hex encoding (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+  // Base64 is chosen to keep compatibility with JSONPb codec.
   repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
 }
 
@@ -81,9 +85,15 @@ message GetOperationsResponse {
 
 service QueryService {
   // GetTraces returns single trace.
+  // Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
+  // and so the response cannot be directly unmarshalled. This can be fixed by standard JSON
+  // parsing libraries or string manipulation to remove the envelope.
   rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
 
   // GetTraces searches for traces.
+  // Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
+  // and so the response cannot be directly unmarshalled. This can be fixed by standard JSON
+  // parsing libraries or string manipulation to remove the envelope.
   rpc GetTraces(FindTracesRequest) returns (stream SpansResponseChunk) {}
 
   // GetServices returns service names.
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -1,5 +1,4 @@
-// Copyright (c) 2019 The Jaeger Authors.
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2021 The Jaeger Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pavolloffay:
@jpkrohling @yurishkuro @joe-elliott could you please review?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,78 @@
...
+
+import "opentelemetry/proto/trace/v1/trace.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "api_v3";
+option java_package = "io.jaegertracing.api_v3";
+
+message GetTraceRequest {
+  bytes trace_id = 1;
```
yurishkuro, line 28: for REST/JSON API, which representation of the trace ID should we support? Jaeger's base16 or OTEL's base64?

jpkrohling: If we expect folks to build tools that can ingest our JSONs as if they were OTLP, we should follow their representation.

pavolloffay: good point. I expect people should be able to copy OTEL traceid (e.g. from logs) and query it directly from Jaeger.

pavolloffay: @yurishkuro where did you find OTEL uses base64? The spec mentions hex encoding. The logging exporters use hex as well.

pavolloffay:
> Is there a reason to provide a completely new set of calls? Instead we could:

> standardize/support the existing HTTP API
add Accept header support for OTLP proto/json to get trace by id/find traces

This is a great question! Technically doable (maybe a bit messier) and provides the same features. One downside is that people will keep using the old API bc the new one is "hidden" behind the accept header.

joe-elliott:
> One downside is that people will keep using the old API bc the new one is "hidden" behind the accept header.
  
This is true. Naturally we could document the list of `Accept` values we support, but I don't disagree it would be harder to find. Is the point of the grpc service to be service to service? Do we expect to migrate the web frontend to use this api (through some kind of http proxy)? Apologies, but I may have missed some discussion about the goals.

proto/api_v3/query_service.proto
yurishkuro: Does OTEL have a spec for JSON format?  If that format is rendered from proto, then it will be base64 for `bytes`

yurishkuro: NB: you can also observe that in Jaeger's JSON from proto, e.g. in the test fixtures. Very annoying (to me), but it's part of the Proto-JSON spec.

jpkrohling: It's experimental, but yes: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp

This is what it says about the trace ID:

> JSON-encoded Protobuf payloads use proto3 standard defined JSON Mapping for mapping between Protobuf and JSON, with one deviation from that mapping: the trace_id and span_id byte arrays are represented as case-insensitive hex-encoded strings, they are not base64-encoded like it is defined in the standard JSON Mapping. The hex encoding is used for trace_id and span_id fields in all OTLP Protobuf messages, e.g. the Span, Link, LogRecord, etc. messages.

yurishkuro: Well, that's a good news, but how is that achieved in OTEL? When I was investigating this same issue for Jaeger JSON model, the only way I could achieve that was by using GoGo's custom marshaller, which wouldn't be too terrible for Go, but it made the .proto definitions useless for all other languages because the off-the-shelf protobuf would encode` bytes` as base64.

pavolloffay: the OTLP receiver uses gogo JSONPb unmarshaller explicitly. I will have a look what other language agents are using.

pavolloffay: Here is the prior discussion on hex vs base64 for ids https://github.com/open-telemetry/opentelemetry-specification/issues/786. PR for making it hex https://github.com/open-telemetry/opentelemetry-specification/pull/911. 

For now I have changed the `trace_id` to be a string type, more problematic are ids in the Span type. The collector does not have marshalling to JSON and the POC was done by using gogo (a similar way how it's done in Jaeger)

yurishkuro: Those OTLP Prs just changed the spec, but didn't discuss how the change was to be implemented when they use proto marshaling.

pavolloffay: I think you mean proto JSON. Look at the issue there is a long discussion with POCs for different languages. Tiagran used gogo and custom type in golang for the POC.

pavolloffay: I have implemented HEX IDs in the linked Jaeger PR - https://github.com/jaegertracing/jaeger/pull/3086. The implementation uses the gogo's custom type feature. I am fine using gogo for now, and then migrate everything once we know what the successor approach will be (e.g. custom marshall or vtprotobuf) 

Here is an example trace https://pastebin.com/J738dzfa from request `curl -ivX GET -H "Content-Type: application/json" http://localhost:16686/v3/traces\?query.serviceName\=frontend`. 

Get trace example `curl -ivX GET -H "Content-Type: application/json" http://localhost:16686/v3/traces/000000000000000061677de41fa1e1e5`

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,78 @@
...
+  string span_kind = 2;
+}
+
+message Operation {
+  string name = 1;
+  string span_kind = 2;
+}
+
+message GetOperationsResponse {
+  repeated Operation operations = 2;
```
albertteoh, line 67: why is the field number 2 and not 1?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2018 Uber Technologies, Inc.
```
albertteoh, line 2: `Copyright (c) 2021 The Jaeger Authors.`

Or perhaps we should start omitting the year [like in OTEL collector](https://github.com/open-telemetry/opentelemetry-collector/blob/c8e28f264c9b293a9bf2b0e7a78c54a962c46650/processor/memorylimiter/internal/cgroups/doc.go#L1)?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,78 @@
...
+
+option go_package = "api_v3";
+option java_package = "io.jaegertracing.api_v3";
+
+message GetTraceRequest {
+  string trace_id = 1;
+}
+
+message SpansResponseChunk {
+  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
```
albertteoh, line 32: what is a resource (maybe [this](https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/StandardResources.md))?

is it possible for jaeger-query to return spans from more than one resource? If so, for my learning, what are some examples?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,78 @@
...
+  string name = 1;
+  string span_kind = 2;
+}
+
+message GetOperationsResponse {
+  repeated Operation operations = 2;
+}
+
+service QueryService {
+  rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
```
joe-elliott, line 71: do you intend to just stream arbitrary batches of spans to the caller as they are found? are these going to be organized by trace?

proto/api_v3/query_service.proto
pavolloffay: good catch, it should be `1`

proto/api_v3/query_service.proto
pavolloffay: OTEL `Resource` is similar to Jaeger `Process` object see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/resource/v1/resource.proto#L27.

Here the returned object is a list of https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L28

proto/api_v3/query_service.proto
pavolloffay: organized by trace. I will add a comment.

pavolloffay:
Added comments and renamed tags to attributes and search depth to num_traces.

proto/api_v3/query_service.proto
yurishkuro: My concern is not how to implement hex in Go - we can use custom marshaller. But if this is to become official API for both grpc and json, then people might want to parse the json payloads using default proto generated classes, and they won't always have the ability to write a custom codec.

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,94 @@
...
+message GetOperationsResponse {
+  repeated Operation operations = 1;
+}
+
+service QueryService {
+  // GetTraces returns single trace.
+  rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
+
+  // GetTraces searches for traces.
+  rpc GetTraces(FindTracesRequest) returns (stream SpansResponseChunk) {}
```
pavolloffay, line 87: @yurishkuro do you remember why streaming was used? An alternative would be to return a list of chunks. Also we could remove chunk to `Trace` which is more idiomatic.

One issue with the streaming and grpc-gateway is that it wraps the response into result - e.g. `{result: {resource_spans: ...}}` see https://github.com/jaegertracing/jaeger/pull/3086/files#diff-1429f7cc5a76981a44799039e43d3bc7372808373b9c2b97a333c7dcf650b00aR72

yurishkuro: Streaming - because large result sets are difficult to transmit as one response.

Returning chunks manually means implementing some kind of pagination API, which would require support in the storage.

proto/api_v3/query_service.proto
pavolloffay: That is a great point. A custom marshaller might be a blocker for people, it's definitely not a one-liner to implement. 

What do you propose then? Use the default base64 for IDs in the objects? For the query parameters I would still prefer to use hex.

yurishkuro: Well, to be honest, this is the reason why I gave up back in the day when trying to make a JSON API backed by proto IDL. I thought OTEL found a solution, but the change in the spec is a total cop out - "it's std proto-JSON except for this field" (which makes std proto-JSON unusable). You mentioned they had prototypes in other languages, how did they solve that?

I am inclined to just support all kinds of formats for IDs in the inputs, i.e. you should be able to paste both base64 and hex ID into the UI. But that doesn't answer what format we return in proto-JSON, and my preference would be to stick with the standard proto-JSON for that, meaning returning base64.

pavolloffay: > You mentioned they had prototypes in other languages, how did they solve that?

custom codes, the difficulty depends on the language so not ideal for the consumers.

proto/api_v3/query_service.proto
pavolloffay: one way or another to fully support streaming the storage API would have to change anyways.

proto/api_v3/query_service.proto
pavolloffay: Also if we go with the streaming API for the get trace(s) the JSONPb codec will not work OOTB - see https://github.com/jaegertracing/jaeger-idl/pull/76#discussion_r652794440. The returned object is wrapped into another object.

pavolloffay: Here is the upstream issue for the reference https://github.com/grpc-ecosystem/grpc-gateway/issues/1254#issuecomment-623634934 and it's apparently not fixed in v2 (I have asked on their slack).

pavolloffay: I have switched impl to use base64 for embedded IDs and keep using hex for queries.

pavolloffay:
@yurishkuro @joe-elliott @albertteoh 

Is there any blocker for this PR? I would like to move it forward. To sum up, the query API exposes OTLP traces. The REST API is done via grpc-gateway with base64 encoding for embedded ids. The ids in the query parameters are hex encoded.

OTEL mandates to use hex encoding for ids in the JSON, however it requires a custom codec which is not easy to implement, hence the result will be hard to consume. To keep the compatibility with JSONPb codec we will expose ids in base64 (alternatively we could have a flag/param to set the encoding).

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+option go_package = "api_v3";
+option java_package = "io.jaegertracing.api_v3";
+
+// Request object to get a trace.
+message GetTraceRequest {
+  // Hex encoded 128 bit trace_id
+  string trace_id = 1;
+}
+
+// A single response chunk holds a single trace.
```
yurishkuro, line 32: In v2 API this description would be inaccurate - a chunk is neither a full trace nor spans from a single trace. A chuck is just a mechanism of delivering large number of spans in smaller batches.

proto/api_v3/query_service.proto
yurishkuro: It's a bit tricky to construct this from Jaeger spans because we denormalize the Process into each individual span. We do have some re-assembly logic when we return spans to the UI, but it's probably also valid to just return (resource, span) pairs as denormalized.

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  // mandates to use hex encoding (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+  // Base64 is chosen to keep compatibility with JSONPb codec.
+  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
+}
+
+// Query parameters to find traces.
+message TraceQueryParameters {
+  string service_name = 1;
+  string operation_name = 2;
+  map<string, string> attributes = 3;
```
yurishkuro, line 48: Should we clarify which attributes these are supposed to match? We've been pretty loose about it in the original API, leaving the interpretation to the storage. I.e. should all these attributes match on a single span, or could they match across spans? Do they match span attributes only or span logs as well?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  google.protobuf.Timestamp start_time_min = 4;
+  google.protobuf.Timestamp start_time_max = 5;
+  google.protobuf.Duration duration_min = 6;
+  google.protobuf.Duration duration_max = 7;
+  int32 num_traces = 8;
```
yurishkuro, line 53: Do we want to keep this as `num_traces` or use a more vague term like "search depth"? Because Cassandra storage does not guarantee num_traces in the response, which was often a source of confusion.

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+
+// Query parameters to find traces.
+message TraceQueryParameters {
+  string service_name = 1;
+  string operation_name = 2;
+  map<string, string> attributes = 3;
+  google.protobuf.Timestamp start_time_min = 4;
+  google.protobuf.Timestamp start_time_max = 5;
+  google.protobuf.Duration duration_min = 6;
+  google.protobuf.Duration duration_max = 7;
```
yurishkuro, line 52: should probably be more explicit about these as well - which spans do these apply to?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  // see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto
+  // In case of JSON format the ids (trace_id, span_id, parent_id) are encoded in base64 even though OpenTelemetry specification
+  // mandates to use hex encoding (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+  // Base64 is chosen to keep compatibility with JSONPb codec.
+  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
+}
+
+// Query parameters to find traces.
+message TraceQueryParameters {
+  string service_name = 1;
```
yurishkuro, line 46: in proto all fields are optional, but we don't support search without service name (although people asked for it). Should we comment that this is required?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+
+// Response object to get service names.
+message GetServicesResponse {
+  repeated string services = 1;
+}
+
+// Request object to search for operation names.
+message GetOperationsRequest {
+  string service = 1;
+  string span_kind = 2;
```
yurishkuro, line 72: comments about required/optional?

proto/api_v3/query_service.proto
yurishkuro: storage API already has `FindTraceIDs`, which allows the query service to load traces one by one

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+service QueryService {
+  // GetTraces returns single trace.
+  // Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
+  // and so the response cannot be directly unmarshalled. This can be fixed by standard JSON
+  // parsing libraries or string manipulation to remove the envelope.
+  rpc GetTrace(GetTraceRequest) returns (stream SpansResponseChunk) {}
+
+  // GetTraces searches for traces.
```
yurishkuro, line 93:
```suggestion
  // FindTraces searches for traces.
```

proto/api_v3/query_service.proto
pavolloffay: Does this hold in practice? 

To my knowledge, a chunk at the moment always represents a single trace (a trace known at the query time).

proto/api_v3/query_service.proto
pavolloffay: The method is called `GetTraces`.

Here is the API mapping 

```
type: google.api.Service
config_version: 3
http:
  rules:
    - selector: jaeger.api_v3.QueryService.GetTrace
      get: /v3/traces/{trace_id}
    - selector: jaeger.api_v3.QueryService.GetTraces
      get: /v3/traces
    - selector: jaeger.api_v3.QueryService.GetServices
      get: /v3/services
    - selector: jaeger.api_v3.QueryService.GetOperations
      get: /v3/operations
```

I didn't use `FindTraces` on purpose to better match the usual REST semantics.

proto/api_v3/query_service.proto
pavolloffay: +1 there are per span, right?

proto/api_v3/query_service.proto
pavolloffay: I would prefer to use `num_traces` it is what the query API uses. If C*  does not implement it (or any other storage) we should document that. 

The goal is to make it clear for users what the parameter means.

proto/api_v3/query_service.proto
pavolloffay: The match should be done on tags and process tags. ES supports match on logs as well when kibana support is enabled (flat schema).

pavolloffay: In ES they must match on a single span. How is it done in cassandra? I think all storages should follow this.

proto/api_v3/query_service.proto
pavolloffay: I have slightly changed the comment

pavolloffay:
PR updated

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,114 @@
...
+import "opentelemetry/proto/trace/v1/trace.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "api_v3";
+option java_package = "io.jaegertracing.api_v3";
+
+// Request object to get a trace.
+message GetTraceRequest {
+  // Hex encoded 128 bit trace_id
```
yurishkuro, line 28: would we not accept 64bit ID here too?

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,114 @@
...
+  string trace_id = 1;
+}
+
+// A single response chunk holds spans from a single trace.
```
yurishkuro, line 32: > To my knowledge, a chunk at the moment always represents a single trace (a trace known at the query time).

I think this is not a matter of how it is currently implemented, but what we want to guarantee.  The intention of the original API was to NOT have this guarantee, i.e. the service is allowed to mix spans from different traces in a single chunk.

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,114 @@
...
+// Note that a subsequent request to get a trace might return a different result
+// because spans are reported in asynchronous chunks.
```
yurishkuro, line 34: > because spans are reported in asynchronous chunks

not sure what this means. There are two reasons why the responses could be different: (1) new spans arrived since the first request, and (2) the order of span loading from storage is not deterministic / guaranteed.

proto/api_v3/query_service.proto
```diff
@@ -0,0 +1,114 @@
...
+message SpansResponseChunk {
+  // OTLP resource spans
+  // see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto
```
yurishkuro, line 37: is this necessary?

If we want to keep the links in the comments I would recommend this format to make the text more readable:
```
comment text [1] more text [2]

[1]: link 1
[2]: link 2
```

proto/api_v3/query_service.proto
yurishkuro: I don't think this is true in Cassandra, because it takes the tag=value string and looks up an index that just gives trace IDs. If more than one tag is provided, it could easily match on different spans.

proto/api_v3/query_service.proto
yurishkuro: what I meant was the same question as with attributes - do we consider the whole TraceQueryParameters clause to be a single conjunction that must be satisfied on a single span?

proto/api_v3/query_service.proto
yurishkuro: * REST endpoints do not need to dictate how gRPC method is called. This will become the default gRPC API
* I would rather have consistency in the API across the backend, everywhere else (e.g. storage API) this is called FindTraces
* should we even have this method that returns full traces, or just have one that returns IDs?
* we do need a method like GetTraces that accepts a collection of trace IDs

pavolloffay: > should we even have this method that returns full traces, or just have one that returns IDs?

proto/api_v3/query_service.proto
pavolloffay: I find it useful to provide the link to the target struct. Not everybody is familiar where OTEL protos are hosted.

proto/api_v3/query_service.proto
pavolloffay: we would, fixed comment.

proto/api_v3/query_service.proto
pavolloffay: I think we should guarantee it for these reasons:
* the current API does not mix spans from different traces
* the API that does not mix spans from different traces is easier to consume (e.g. for typical use-cases that we know)

proto/api_v3/query_service.proto
pavolloffay: maybe it's better to remove this comment

proto/api_v3/query_service.proto
pavolloffay: I have added a comment on the `TraceQueryParameters` that clarifies that some storage implementations might deviate.

Alright so C* does deviate as well. What was the original design for attributes? Match any attributes within trace or in a single span?

jpkrohling: Another way to see this: what would you expect as a user? If you specify two pairs of attributes, would you expect them to exist throughout the trace, or for all attributes to exist as part of the same span? I'm not quite sure I have an answer here... Here's one case advocating for attributes to exist throughout the trace:

- root span has "userID=123", no other spans contain this
- span (not root span) has error=true
- SRE is looking for traces for user 123 with errors in it

proto/api_v3/query_service.proto
jpkrohling: This is a common source of confusion for our users. Most users I know expect the duration filter to apply to the trace, not to spans within the trace.

proto/api_v3/query_service_http.yaml
```diff
@@ -0,0 +1,14 @@
+# This is an API configuration to generate an HTTP/JSON -> gRPC gateway for the
+# OpenTelemetry service using github.com/grpc-ecosystem/grpc-gateway.
```
jpkrohling, line 2:
```suggestion
# Query service using github.com/grpc-ecosystem/grpc-gateway.
```

proto/api_v3/query_service.proto
pavolloffay: There is a comment which should make it clear that it is for `Span` 

> // Span min duration.

pavolloffay:
@yurishkuro are there any blockers on your side for this PR? 

The biggest shady point in the PR is the definition of `TraceQueryParameters` parameters. I have added a comment 

> // Note that some storage implementations do not guarantee the correct implementation of all parameters.

proto/api_v3/query_service.proto
yurishkuro: I still don't like this guarantee in the API. The streaming & chunking API was primarily introduced for efficiency, but we're taking away server's ability to optimize its response. You could be loading a ton of small traces, e.g. 2 spans each, so this guarantee in the API would force the server to send tiny chunks, which is going to be suboptimal. On the other hand, there is a max chunk size in the server so there is always a possibility that a large trace will be split across several chunks, that largely takes away your "easier to consume" reason #2.

We can always introduce this guarantee later, but removing it would backwards incompatible.

proto/api_v3/query_service.proto
jpkrohling: Doesn't make sense to fix this behavior for this new API? We can claim it works as documented, but users may claim it doesn't work as expected.

proto/api_v3/query_service.proto
pavolloffay: In the 5 years of the project history we haven't used the feature of streaming chunks with mixed traces. I don't think any other DT system behaves this way.

>  You could be loading a ton of small traces, e.g. 2 spans each, so this guarantee in the API would force the server to send tiny chunks, which is going to be suboptimal.

This is how Jaeger works right now and we haven't see any complaints/issues/use-case to change it.

pavolloffay: > We can always introduce this guarantee later, but removing it would backwards incompatible.

It would be a breaking change one way or another. The current consumers do not expect that spans in chunks are mixed (e.g. UI does not).

proto/api_v3/query_service.proto
pavolloffay: I agree that the current UI search is confusing - there is duration but it does not say whether it is per span or per trace.

The current implementation uses span duration. Span duration is well defined (a span has start and end time) and can be directly queried from the storage. 

Trace duration is more complicated. It's not clearly defined and cannot be queried from the storage. The naive implementation could take the first and the last timestamp and the time in between consider as a trace duration, but is that correct? It does not consider async and messaging scenarios. It is also a breaking change to the current behavior.

jpkrohling: As a user, what is most useful? Searching for the duration of individual spans, or for the duration of a trace? If you think it's span duration, I'm in having it this way.

proto/api_v3/query_service.proto
pavolloffay: I don't agree with this per my comments above but I have removed the guarantee of not mixing spans in one chunk per Yuri's request.

proto/api_v3/query_service.proto
pavolloffay: There won't be a clear answer, both are useful. I would certainly use span duration when querying for slow traces for a given API or DB call.

When I don't have narrow search space I would like to get the slow transactions per service. This is how the UI behaves the search traces results can be sorted per trace duration.

In most cases, the root span represents trace duration. In async messaging scenarios it's not the case but I would as well argue that trace duration in those cases is not a proper duration (e.g. is the duration result of endTime-startTime or sum of span durations). Maybe in these cases users would like to sort traces based on the end time?

yurishkuro (approved):

proto/api_v3/query_service.proto
jpkrohling: If you want to create an issue to address this, this PR can be merged as is.

PR #77, opened 2021-06-28
Fix selector for grpc gateway in apiv3

pavolloffay:
This is a bug fix, without this the compilation for grp-gateway does not work

### DIFF
--- a/proto/api_v3/query_service_http.yaml
+++ b/proto/api_v3/query_service_http.yaml
@@ -6,7 +6,7 @@ http:
   rules:
     - selector: jaeger.api_v3.QueryService.GetTrace
       get: /v3/traces/{trace_id}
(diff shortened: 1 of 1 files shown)


proto/api_v3/query_service_http.yaml
```diff
@@ -6,7 +6,7 @@ http:
   rules:
     - selector: jaeger.api_v3.QueryService.GetTrace
       get: /v3/traces/{trace_id}
-    - selector: jaeger.api_v3.QueryService.GetTraces
+    - selector: jaeger.api_v3.QueryService.FindTraces
```
jpkrohling, line 9: Is there a way to test this?

jpkrohling (approved):
LGTM, but a good follow-up task would be to add a test to the CI.

PR #80, opened 2021-07-01
Document required fields in api_v3

pavolloffay:
Example query: 
```
curl -H "Content-Type: application/json" http://localhost:16686/v3/traces\?query.serviceName\=jaeger-query\&query.start_time_min\="2021-07-01T01:00:00Z"\&query.start_time_max\=2021-07-03T01:00:00Z\&query.duration_min\=0\&query.duration_max\=10s\&query.num_traces\=150000
```

Documenting this as it took me a while to realize which parameters are required and what is the format.

### DIFF
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -48,13 +48,13 @@ message TraceQueryParameters {
   // Attributes are matched against Span and Resource attributes.
   // At least one span in a trace must match all specified attributes.
   map<string, string> attributes = 3;
-  // Span min start time in. RFC-3339ns format. Required.
+  // Span min start time in. REST API uses RFC-3339ns format. Required.
   google.protobuf.Timestamp start_time_min = 4;
-  // Span max start time. RFC-3339ns format. Required.
+  // Span max start time. REST API uses RFC-3339ns format. Required.
   google.protobuf.Timestamp start_time_max = 5;
-  // Span min duration. Golang duration format. Required.
+  // Span min duration. REST API uses Golang's time format e.g. 10s. Required.
   google.protobuf.Duration duration_min = 6;
-  // Span max duration. Golang duration format. Required.
+  // Span max duration. REST API uses Golang's time format 10s. Required.
   google.protobuf.Duration duration_max = 7;
   // Maximum number of traces in the response.
   int32 num_traces = 8;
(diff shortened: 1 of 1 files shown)


proto/api_v3/query_service.proto
```diff
@@ -48,13 +48,13 @@ message TraceQueryParameters {
...
   // At least one span in a trace must match all specified attributes.
   map<string, string> attributes = 3;
-  // Span min start time.
+  // Span min start time in. RFC-3339ns format. Required.
   google.protobuf.Timestamp start_time_min = 4;
-  // Span max start time.
+  // Span max start time. RFC-3339ns format. Required.
   google.protobuf.Timestamp start_time_max = 5;
-  // Span min duration.
+  // Span min duration. Golang duration format. Required.
```
yurishkuro, line 55: Duration was not a required param previously.

> Golang duration format.

Not sure what it means in proto definition, the format is defined by google.protobuf.Duration, not by Golang.

proto/api_v3/query_service.proto
```diff
@@ -48,13 +48,13 @@ message TraceQueryParameters {
   // Attributes are matched against Span and Resource attributes.
   // At least one span in a trace must match all specified attributes.
   map<string, string> attributes = 3;
-  // Span min start time.
+  // Span min start time in. RFC-3339ns format. Required.
```
yurishkuro, line 51: > RFC-3339ns

A link would be useful, but not sure what "format" means in this case in proto, since it's determined by google.protobuf.Timestamp definition.

yurishkuro: If you're referring to the format in JSON, then we should phrase it differently. Doesn't grpc-gateway allow generating swagger definition? Ideally we should channel these types of comments there.

pavolloffay: > Doesn't grpc-gateway allow generating swagger definition? Ideally we should channel these types of comments there.

I didn't find annotations for this.

proto/api_v3/query_service.proto
pavolloffay: for the REST API it is golang's time format. Can you tell that by looking at https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration?

yurishkuro: that's probably the grpc-gateway's interpretation? In that link many types have a dedicated JSON section, but Duration does not, and since it's just a composite proto type I think its default JSON representation would be a struct, not a Go duration string, so it must be somehow handled differently in the translation. The same probably applies to the Timestamp type.

pavolloffay: I don't know, but suing go time strings work. Hence I have documentedthat.

pavolloffay:
Duration should not be required. I have changed it.

yurishkuro (approved):

PR #81, opened 2021-07-02
Check in swagger for v3 query service

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

I think this will be useful for people who want to use the REST API. Initially I have checked this file in the implementation in Jaeger alongside compiled protos, but it makes more sense to have it here.

### DIFF
--- a/swagger/api_v3/query_service.swagger.json
+++ b/swagger/api_v3/query_service.swagger.json
@@ -1,0 +1,657 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "api_v3/query_service.proto",
+    "version": "version not set"
+  },
+  "consumes": [
+    "application/json"
+  ],
+  "produces": [
+    "application/json"
+  ],
+  "paths": {
+    "/v3/operations": {
+      "get": {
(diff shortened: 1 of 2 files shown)


swagger/api_v3/query_service.swagger.json
```diff
@@ -0,0 +1,657 @@
+{
```
pavolloffay, line 1: note that we already host zipkin swagger that is in yaml format https://github.com/jaegertracing/jaeger-idl/tree/master/swagger. 

the protoc cannot generate yaml, we could translate the file using `yq`, I do not fancy it very much.

jpkrohling (approved):

swagger/api_v3/query_service.swagger.json
```diff
@@ -0,0 +1,657 @@
...
+              "$ref": "#/definitions/api_v3GetOperationsResponse"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "service",
+            "description": "Required service name.",
+            "in": "query",
+            "required": false,
```
yurishkuro, line 31: Room for improvement

PR #87, opened 2021-12-14
Document how strategyType field is used

### LINKED ISSUE #86: Is SamplingStrategyResponse.strategyType field being used?
`SamplingStrategyResponse` can contain one of 3 types of sampling strategies, but the `SamplingStrategyType` enum only defines 2 values.

The `strategyType` field is not used by the Go client: https://github.com/jaegertracing/jaeger-client-go/blob/master/sampler_remote.go

Need to check other langs.

yurishkuro:
Resolves #86 

Closes #88

### DIFF
--- a/proto/api_v2/sampling.proto
+++ b/proto/api_v2/sampling.proto
@@ -84,7 +84,7 @@ message PerOperationSamplingStrategies {
 }
 
 // SamplingStrategyResponse contains an overall sampling strategy for a given service.
-// This type should be considered a unit where only one of the strategy field is present.
+// This type should be treated as a union where only one of the strategy field is present.
 message SamplingStrategyResponse {
   // Legacy field that was meant to indicate which one of the strategy fields
   // below is present. This enum was not extended when per-operation strategy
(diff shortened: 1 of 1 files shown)


proto/api_v2/sampling.proto
```diff
@@ -57,10 +58,21 @@ message PerOperationSamplingStrategies {
 }
 
 message SamplingStrategyResponse {
+  // Legacy field that was meant to indicate which one of the strategy fields
+  // below is present. This enum was not extended when per-operation strategy
+  // was introduced, because extending enum has backwards compatiblity issues.
+  // The recommended approach for consumers is to ignore this field and instead
+  // checks the other fields being not null (starting with operationSampling).
```
pavolloffay, line 65: >  checks the other fields being not null

IIRC in proto3 fields are not nullable https://stackoverflow.com/a/42634681.

We should make it clear what is the precedence order for different samplers.

yurishkuro: That SO answer is nonsense. Proto encodes structure by listing the field number in the payload. If the field number is not present it means it is null. All grpc-generated classes use pointers for complex fields, making null checks trivial.

pavolloffay: >  Proto encodes structure by listing the field number in the payload. If the field number is not present it means it is null

I get the encoding specification, but if standard `protoc` generator is used (which is used in most cases) how is the null checked?

pavolloffay: nevermind there is `response.hasOperationSampling()`. I got confused bc the `response.getOperationSampling()` returns default instance if it is null.

yurishkuro: I haven't worked with protoc in all languages, but for Go it would generate pointer fields, which will be set to null if the field is empty. We specifically had to work around that by using gogoproto to force generated types to inline/nest other types instead of pointers. I expect that the same would happen with other langs.

> hasOperationSampling()

right, that too

PR #88, opened 2021-12-15
Add docs for sampling

pavolloffay:
@yurishkuro could you please review? 

This was originally added here https://github.com/open-telemetry/opentelemetry-specification/pull/2222 but moving here as the spec will reference this file.

### DIFF
--- a/proto/api_v2/sampling.proto
+++ b/proto/api_v2/sampling.proto
@@ -36,26 +36,39 @@ enum SamplingStrategyType {
   RATE_LIMITING = 1;
 };
 
+// ProbabilisticSamplingStrategy defines TraceIdRatioBased sampler configuration.
 message ProbabilisticSamplingStrategy {
+  // // Sampling rate between [0-1].
   double samplingRate = 1;
 }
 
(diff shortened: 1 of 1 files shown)


proto/api_v2/sampling.proto
```diff
@@ -36,37 +36,53 @@ enum SamplingStrategyType {
...
 message OperationSamplingStrategy {
+  // Operation name is a span name.
   string operation = 1;
+  // Probabilistic sampler configuration.
   ProbabilisticSamplingStrategy probabilisticSampling = 2;
 }
 
+// PerOperationSamplingStrategies defines sampling configuration for operations/span (names).
 message PerOperationSamplingStrategies {
+  // Default sampling probability for TraceIdRatioBased sampler.
```
pavolloffay, line 61: @yurishkuro I think comments in this type need more explanation. 

The per-operation config is clear - it is used if names match, if name does not match does it fallback to the probabilistic and rate limiting sampler? Is the rate limiting sampler used first until a specified throughput is satisfied and then it is switched to probabilistic?

yurishkuro (commented):
Let me first merge the other pr about the strategyType and I will write proper docs for the remaining types.

PR #89, opened 2022-02-02
Update query_service.proto

feldentm:
Explain service_name trace query parameter

- lack of documentation

- add documentation reflecting the current behaviour

### DIFF
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -43,6 +43,7 @@ message SpansResponseChunk {
 // Query parameters to find traces.
 // Note that some storage implementations do not guarantee the correct implementation of all parameters.
 message TraceQueryParameters {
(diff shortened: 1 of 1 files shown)


proto/api_v3/query_service.proto
```diff
@@ -43,6 +43,7 @@ message SpansResponseChunk {
 // Query parameters to find traces.
 // Note that some storage implementations do not guarantee the correct implementation of all parameters.
 message TraceQueryParameters {
+  // Service name as shown in the frontend. Required.
```
yurishkuro, line 46: I don't think "in the frontend" is a useful definition. Service Name is defined in the data model: `Span.Process.serviceName`

feldentm: Fine for me. I'd just prefer to have some details here :)

PR #92, opened 2022-03-08
swagger: fix spelling mistakes in zipkin2-api.yaml

kevinburkesegment:


yurishkuro:
The zipkin files are copied from the Zipkin project, we're not going to change them.

PR #95, opened 2023-04-11
Add start and end timestamp args to GetTrace request

alburthoffman:
- https://github.com/jaegertracing/jaeger/issues/4150

- allow optional time parameters when querying trace by id

### DIFF
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -27,10 +27,10 @@ option java_package = "io.jaegertracing.api_v3";
 message GetTraceRequest {
   // Hex encoded 64 or 128 bit trace ID.
   string trace_id = 1;
-  // Optional. The start time to search trace id
-  google.protobuf.Timestamp start = 2;
-  // Optional. The end time to search trace id
-  google.protobuf.Timestamp end = 3;
+  // Optional. The start time to search trace ID.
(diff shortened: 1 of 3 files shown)


yurishkuro (approved):

yurishkuro:
Please make sure commits are signed.

alburthoffman:
sure. something wrong with my local env. will fix and sign the commit.

yurishkuro (approved):

proto/api_v3/query_service.proto
```diff
@@ -27,6 +27,10 @@ option java_package = "io.jaegertracing.api_v3";
 message GetTraceRequest {
   // Hex encoded 64 or 128 bit trace ID.
   string trace_id = 1;
+  // Optional. The start time to search trace id
+  optional google.protobuf.Timestamp start = 2;
+  // Optional. The end time to search trace id
+  optional google.protobuf.Timestamp end = 3;
```
yurishkuro, line 33: CI fails on the `optional` keyword. IIRC all fields are optional by default, so let's drop the keyword.

yurishkuro (approved):

yurishkuro:
Do you only need this change in V3? Your original ticket mentioned Grafana as UI, does it query Jaeger using V3 API (OTLP-based)?

PR #107, opened 2024-10-10
Add a more flexible sampling strategy data model

garrettlish:
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6127

## Description of the changes
- Add a more flexible sampling strategy data model, it enables support for more fine-grained control, such as adjusting sampling probabilities based on specific tag key-value pairs (see [sampling.proto](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/samplers/jaegerremote/jaeger-idl/proto/api_v2/sampling.proto)).
- For instance, in a real-world scenario, we might want to enforce sampling for a particular user session by customizing sampling probabilities based on specific tag key-value pairs.

- Tested in the upcoming PR.

### DIFF
--- a/proto/api_v2/sampling.proto
+++ b/proto/api_v2/sampling.proto
@@ -83,44 +83,67 @@ message PerOperationSamplingStrategies {
   double defaultUpperBoundTracesPerSecond = 4;
 }
 
-// DimensionMatcher is a key-value pair that defines a dimension and a matching strategy.
-message DimensionMatcher {
-  enum Type {
-    EQ = 0; // Equals
-    NEQ = 1; // Not equals
-    RE = 2; // Regular expression match
-    NRE = 3; // Not regular expression match
-  }
-  Type type = 1;
-  string key = 2;
-  string value = 3;
-}
-
-// MultiDimensionalSamplingStrategy is a sampling strategy that takes into account arbitrary
-// combinations of dimensions (aka user session). Only probabilistic sampling is currently supported.
-message MultiDimensionalSamplingStrategy {
-  // Whether to match any or all conditions.
-  bool anyOrConjunction = 1;
-
-  // List of dimension matchers.
-  repeated DimensionMatcher matchers = 2;
+// RuleBasedSamplingStrategy defines rule-based sampling strategy.
+message RuleBasedSamplingStrategy {
+  // Logical expression combining conditions
+  LogicalExpression expression = 1;
 
   // Probabilistic sampling strategy.
-  ProbabilisticSamplingStrategy probabilisticSampling = 3;
+  ProbabilisticSamplingStrategy probabilisticSampling = 2;
 }
 
-// PerMultiDimensionalSamplingStrategy is a combination of strategies for different dimensions
-// as well as some service-wide defaults. It is particularly useful for services whose
-// dimensions receive vastly different traffic, so that any single rate of sampling would
-// result in either too much data for some dimensions or almost no data for other dimensions.
-message PerMultiDimensionalSamplingStrategy {
+// LogicalExpression represents a logical expression which could be a single condition
+// or a combination of conditions.
+message LogicalExpression {
+  oneof expression {
+    Condition condition = 1;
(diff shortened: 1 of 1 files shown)


yurishkuro:
This feels too simplistic to me. Compare with https://github.com/open-telemetry/oteps/pull/250

garrettlish:
> This feels too simplistic to me. Compare with [open-telemetry/oteps#250](https://github.com/open-telemetry/oteps/pull/250)

Good idea. How about introducing an "any" or "conjunction" option, along with adding a dimension matching strategy, similar to the changes made in this commit - https://github.com/jaegertracing/jaeger-idl/pull/107/commits/171bccd3cd294806364c6bf69b0545eb15a7adab? Any thoughts, please advise!

yurishkuro:
I would recommend starting with some use cases and showing a yaml-equivalent of the config for those, then we can translate it to the IDL. I think your proposal is still too rigid. For example, here's the DLS that Gemini recommends:
```proto
message SamplingRule {
  // Logical expression combining conditions
  LogicalExpression expression = 1;

  // Sampling strategy
  oneof sampling_strategy {
    FixedRateSampling fixed_rate_sampling = 2;
    ProbabilisticSampling probabilistic_sampling = 3;
    // Add other sampling strategies as needed
  }
}

message LogicalExpression {
  oneof expression {
    Condition condition = 1;
    AndExpression and_expression = 2;
    OrExpression or_expression = 3;
    NotExpression not_expression = 4;
  }
}

message AndExpression {
  repeated LogicalExpression expressions = 1;
}

message OrExpression {
  repeated LogicalExpression expressions = 1;
}

message NotExpression {
  LogicalExpression expression = 1;
}
```

example rule
```proto
message SamplingRule {
  expression {
    and_expression {
      expressions {
        condition {
          attribute: "http.status_code"
          operator: GREATER_THAN
          int_value: 500
        }
      }
      expressions {
        condition {
          attribute: "service.name"
          operator: EQUALS
          string_value: "my_service"
        }
      }
    }
  }
  fixed_rate_sampling {
    rate: 0.5
  }
}
```
The challenge when dealing with expressions and boolean logic is in validating that the rules in a strategy are disjoint, as otherwise the stratas are going to overlap and proper stratified sampling probabilities cannot be determined (although we probably can deal with that by assuming a priority order of the rules where earlier rule wins, in which case even if two rules overlap in the search space prioritizing one of them will effectively make them disjoint).

proto/api_v2/sampling.proto
```diff
@@ -83,6 +83,69 @@ message PerOperationSamplingStrategies {
...
+enum Operator {
+  EQUALS = 0;
+  GREATER_THAN = 1;
+  LESS_THAN = 2;
+  REGEX_MATCH = 3;
+}
+
+// PerRuleBasedSamplingStrategy defines the default sampling probability and multiple
+// rule-based sampling strategies.
+message PerRuleBasedSamplingStrategy {
```
yurishkuro, line 139:
```suggestion
message RuleBasedSamplingStrategy {
```

proto/api_v2/sampling.proto
```diff
@@ -83,6 +83,69 @@ message PerOperationSamplingStrategies {
   double defaultUpperBoundTracesPerSecond = 4;
 }
 
+// RuleBasedSamplingStrategy defines rule-based sampling strategy.
+message RuleBasedSamplingStrategy {
```
yurishkuro, line 87:
```suggestion
message SamplingRule {
```

proto/api_v2/sampling.proto
```diff
@@ -83,6 +83,69 @@ message PerOperationSamplingStrategies {
...
+  // Logical expression combining conditions
+  LogicalExpression expression = 1;
```
yurishkuro, line 89:
```suggestion
  LogicalExpression condition = 1;
```

yurishkuro (commented):
I noticed I never submitted pending review, sorry about that.

If you are not interested in continuing with this please close the PR.

PR #114, opened 2024-12-26
[query][api_v2/v_3] Add raw_traces parameter to GetTraceRequest and FindTracesRequest

mahadzaryab1:
- Towards https://github.com/jaegertracing/jaeger/issues/6417

## Description of the changes
- Added the `raw_traces` parameter to the `GetTraceRequest` and the `TraceQueryParameters` in the `FindTracesRequest`

## Checklist

### DIFF
--- a/proto/api_v2/query.proto
+++ b/proto/api_v2/query.proto
@@ -50,6 +50,9 @@ message GetTraceRequest {
     (gogoproto.stdtime) = true,
     (gogoproto.nullable) = false
   ];
+  // Optional. If set to true, the response will not perform any 
+  // adjustments to the trace. 
+  bool raw_traces = 4;
 }
 
 message SpansResponseChunk {
@@ -112,6 +115,7 @@ message TraceQueryParameters {
     (gogoproto.nullable) = false
   ];
   int32 search_depth = 8;
+  bool raw_traces = 9;
 }
 
 message FindTracesRequest {
--- a/proto/api_v2/query.proto
+++ b/proto/api_v2/query.proto
@@ -116,6 +116,9 @@ message TraceQueryParameters {
     (gogoproto.nullable) = false
   ];
   int32 search_depth = 8;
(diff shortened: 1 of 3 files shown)


proto/api_v2/query.proto
```diff
@@ -112,6 +115,7 @@ message TraceQueryParameters {
     (gogoproto.nullable) = false
   ];
   int32 search_depth = 8;
+  bool raw_traces = 9;
```
yurishkuro, line 118: Same for Find?

mahadzaryab1: @yurishkuro this is for `FindTraces`

proto/api_v2/query.proto
```diff
@@ -50,6 +50,9 @@ message GetTraceRequest {
     (gogoproto.stdtime) = true,
     (gogoproto.nullable) = false
   ];
+  // Optional. If set to true, the response will not perform any 
+  // adjustments to the trace.
```
yurishkuro, line 54:
```suggestion
  // enrichments to the trace, such as clock skew adjustment.
  // Instead, the trace will be returned exactly as stored.
```

proto/api_v2/query.proto
yurishkuro: oh - let's add the same documentation to the field.

Also, why are you limiting this to api_v2 only? Let's make the same change to api_v3

mahadzaryab1: @yurishkuro I was going to do v3 in a separate PR but let me know if you prefer it in this one.

mahadzaryab1: https://github.com/jaegertracing/jaeger-idl/pull/115

yurishkuro: I would do both in one

yurishkuro (approved):

PR #116, opened 2025-01-14
Add Go code generated for api_v2

adityachopra29:
- Part 1 and 2 of https://github.com/jaegertracing/jaeger/issues/6494 for api_v2

## Description of the changes
- Create makefile for compiling and generating code from .proto files present
- All locations of proto files, as well as generated code are inspired from main jaeger repo. (Since no other alteranative was suggested in issue.)
- Add all the generated code from the proto files as well.
- Create ci-lint file for same.

- CI

## Checklist
- [ ] I have added unit tests for the new functionality
- [ ] I have run lint and test steps successfully


adityachopra29:
@yurishkuro  Please have a look now. 
I was not sure where all required proto files must be added, so i took inspiration from the main repo, can change it if that is not required.

internal/proto/api_v3/query_service.pb.go
```diff
@@ -0,0 +1,3357 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
yurishkuro, line 1: it won't be usable if it's in the internal package.

plugin/storage/grpc/proto/storage_test.proto
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2019 The Jaeger Authors
```
yurishkuro, line 1: this is not useful here, we can leave it in Jaeger

model/prototest/model_test.pb.go
```diff
@@ -0,0 +1,230 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
yurishkuro, line 1: not needed in this repo

yurishkuro (commented):
I would recommend skipping /model/ for now, because it has a much higher knock-off effect on Jaeger repo, while the other types are easier to swap, and do not require much of organization. The original ticket was not asking to migrate /model/.

Please ensure that all generated files are organized into a consistent folder hierarchy, rooted at `proto-gen`

I would recommend starting with a single .proto, e.g. api_v2. There is more housekeeping work that needs to be done first, such as setting up CI to lint & compile, defining a release process, etc. Let's do one module e2e before migrating everything.

adityachopra29:
> I would recommend starting with a single .proto, e.g. api_v2. There is more housekeeping work that needs to be done first, such as setting up CI to lint & compile, defining a release process, etc. Let's do one module e2e before migrating everything.

Okay. I was trying to understand how to setup CI, etc. Can you please explain what all is required for setting up the CI?

yurishkuro:
what we want in the CI:
  * a re-run of the code generation and fail CI if it finds a difference (see how it's done in the jaeger repo)
  * I think that's it for now

Makefile.Protobuf.mk
```diff
@@ -0,0 +1,85 @@
...
+
+  $(PROTOC) \
+    $(PROTO_INCLUDES) \
+    --gogo_out=plugins=grpc,$(strip $(4)),$(PROTO_GOGO_MAPPINGS):$(PWD)/$(strip $(1)) \
+    $(3) $(2)
+
+endef
+
+.PHONY: new-proto
+new-proto: new-proto-api-v2
```
yurishkuro, line 78: Why "new"?

adityachopra29: @yurishkuro 
There is already a command ```make proto``` in the main Makefile. I didn't want to remove that right now before confirming, so I made the ```make new-proto``` command in the meantime.
If we can remove the old make proto command, we can rename it.

yurishkuro: why do we need two different paths at all? The difference between the two commands is that the old one generates for all languages (and is used mostly as a linter step), while the new one only generates Go types. I would prefer to avoid duplication and be able to support both Go and All languages with a single set of commands.

yurishkuro: you know, let's hold off on this topic - we can refactor this as as separate issue, I would like to get api_v2 types into this repo first.

yurishkuro:
I am seeing CI failures
```
	proto/api_v3/query_service.proto
opentelemetry/proto/trace/v1/trace.proto: File not found.
proto/api_v3/query_service.proto:19:1: Import "opentelemetry/proto/trace/v1/trace.proto" was not found or had errors.
proto/api_v3/query_service.proto:165:3: "opentelemetry.proto.trace.v1.TracesData" is not defined.
proto/api_v3/query_service.proto:125:49: "opentelemetry.proto.trace.v1.TracesData" is not defined.
proto/api_v3/query_service.proto:129:53: "opentelemetry.proto.trace.v1.TracesData" is not defined.
make: *** [Makefile:153: proto-api-v3] Error 1
```

why is the original workflow affected, what changed? Does your new makefile override something (like variables) in the old one?

Makefile.Protobuf.mk
adityachopra29: > why do we need two different paths at all?

Yes, I agree. I had planned on replacing that test in the end. Just wanted to show you the changes before making them; hence, I made a new file.

Okay, I will create another pr to make these changes

adityachopra29:
@yurishkuro  My bad, I had redefined the PROTO_INCLUDES variable, because of which the error was coming. Have pushed the changes

Along the way, I also saw some command variables that were hardcoded in the Makefile (such as docker, sed). We can also change them in another pr.

Makefile.Protobuf.mk
```diff
@@ -0,0 +1,70 @@
...
+# --gogo_out generates GoGo Protobuf output with gRPC plugin enabled.
+# --govalidators_out generates Go validation files for our messages types, if specified.
+#
+# The lines starting with Mgoogle/... are proto import replacements,
+# which cause the generated file to import the specified packages
+# instead of the go_package's declared by the imported protof files.
+#
+
+DOCKER_PROTOBUF_VERSION=0.5.0
+DOCKER_PROTOBUF=jaegertracing/protobuf:$(DOCKER_PROTOBUF_VERSION)
```
yurishkuro, line 18: there are already vars defined in the main Makefile for the same concept, should be using them.

yurishkuro (approved):

yurishkuro:
As discussed, I am going to merge this so that we can start using these types in the jaeger repo. A clean-up / deduplication can be done separately.

yurishkuro:
@adityachopra29 we will also need go.mod somewhere to make this usable as a go library. And it would be good to have some minimal test to ensure the generated code compiles.

PR #120, opened 2025-01-20
Clean Makefile

adityachopra29:
- Cleans the Makefile as discussed here: https://github.com/jaegertracing/jaeger-idl/pull/116#discussion_r1919052873

## Description of the changes
- Removes the redundant Makefile.Protobuf.mk file

## Checklist
- [ ] I have added unit tests for the new functionality
- [ ] I have run lint and test steps successfully


Makefile
```diff
@@ -71,7 +71,8 @@ PROTO_GOGO_MAPPINGS := $(shell echo \
 		Mmodel.proto=github.com/jaegertracing/jaeger-idl/model/v1 \
 	| sed 's/ //g')
 
-PROTO_GEN_GO_DIR ?= proto-gen-go
+PROTO_GEN_GO_DIR ?= proto-gen
+PROTO_GEN_GO_DIR_OLD ?= proto-gen-go
```
yurishkuro, line 75: It's not "old", we are not planning to remove it, it's just "all languages".
  * you can name it `POLYGLOT`
  * I would prefer putting this stuff into a dot-directory, e.g. `.proto-gen-polyglot`, right now if you run this target in your clone it pollutes the root directory with all kinds of garbage. There's also another set of dis that start with just `gen-*`

This can be a separate PR.

Makefile
```diff
@@ -83,9 +84,8 @@ PROTO_GEN_CSHARP_DIR ?= proto-gen-csharp
 
 PROTOC_WITHOUT_GRPC_common := $(PROTOC) \
 		$(PROTO_INCLUDES) \
-		--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/${PROTO_GEN_GO_DIR} \
 		--python_out=${PROTO_GEN_PYTHON_DIR} \
-		--js_out=${PROTO_GEN_JS_DIR}
+		--js_out=${PROTO_GEN_JS_DIR} \
```
yurishkuro, line 88: what's happening here?

go.mod
```diff
@@ -7,6 +7,8 @@ toolchain go1.23.5
 require (
 	github.com/gogo/googleapis v1.4.1
 	github.com/gogo/protobuf v1.3.2
+	github.com/golang/protobuf v1.5.4
+	github.com/grpc-ecosystem/grpc-gateway v1.16.0
```
yurishkuro, line 11: why is this changing? Did you run `go mod tidy` after generating all languages? This is why I want polyglot to be in a dot-directory, it's not a committed source code.

Makefile
```diff
@@ -139,6 +153,7 @@ proto: proto-prepare proto-api-v2 proto-api-v3
 .PHONY: proto-prepare
 proto-prepare:
 	mkdir -p ${PROTO_GEN_GO_DIR} \
+		${PROTO_GEN_GO_DIR_OLD} \
 		${PROTO_GEN_JAVA_DIR} \
```
yurishkuro, line 157: all these language dirs should be nested inside dot-directory

Makefile
```diff
@@ -155,6 +170,12 @@ proto-api-v2:
 		proto/api_v2/collector.proto \
 		proto/api_v2/sampling.proto
 
+	mkdir -p ${PROTO_GEN_GO_DIR}/api_v2
```
yurishkuro, line 173: I don't want to combine generation of committed code with generation of uncommitted code. Please use separate targets. `make proto` should only generated committed code. `make proto-all` can generate all languages.

Makefile
adityachopra29: Okay👍 Will do right after this.

Makefile
adityachopra29: My bad. I was changing the path while experimenting with something and this go changed. Will change this.

go.mod
adityachopra29: Yes. That was also the reason I had run `go get google.golang.org/genproto@latest`, since `go mod tidy` was giving the error I had referenced in the description. I got it now.

Makefile
adityachopra29: Okay

Makefile
adityachopra29: Also, I had removed the 
`--gogo_out=plugins=grpc,$(PROTO_GOGO_MAPPINGS):$(PWD)/${PROTO_GEN_GO_DIR} \` because I thought that we will be removing the generated code for all languages. I will put this back as well

yurishkuro: the generation for all languages is a sanity check that our .proto files can be compiled, if people want to build their own API clients.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/
 gen-*/
 proto-gen-*/
```
yurishkuro, line 3: remove this?

Makefile
```diff
@@ -126,27 +131,56 @@ else
...
+    --gogo_out=plugins=grpc,$(strip $(4)),$(PROTO_GOGO_MAPPINGS):$(PWD)/$(strip $(1)) \
+    $(3) $(2)
+
+endef
 
 .PHONY: test-ci
 test-ci:
 	go test -v -coverprofile=coverage.txt ./...
 
 .PHONY: proto
```
yurishkuro, line 155:
```suggestion
# proto target is used to generate source code that is released as part of this library
.PHONY: proto
```

Makefile
```diff
@@ -126,27 +131,56 @@ else
...
-proto: proto-prepare proto-api-v2 proto-api-v3
+proto: proto-prepare proto-api-v2
+
+.PHONY: proto-all
```
yurishkuro, line 158:
```suggestion
# proto-all target is used to generate code for all languages as a validation step.
# The resulting code is not committed.
.PHONY: proto-all
```

Makefile
```diff
@@ -126,27 +131,56 @@ else
...
+proto-prepare-all:
+	mkdir -p ${PROTO_GEN_GO_DIR_POLYGLOT} \
+		${PROTO_GEN_JAVA_DIR_POLYGLOT} \
+		${PROTO_GEN_PYTHON_DIR_POLYGLOT} \
+		${PROTO_GEN_JS_DIR_POLYGLOT} \
+		${PROTO_GEN_CPP_DIR_POLYGLOT} \
+		${PROTO_GEN_CSHARP_DIR_POLYGLOT}
 
 .PHONY: proto-prepare
 proto-prepare:
```
yurishkuro, line 171: group real vs. transient generation targets together, right now they are mixed

Makefile
```diff
@@ -126,27 +131,56 @@ else
...
-include Makefile.Protobuf.mk
+# Macro to compile Protobuf $(2) into directory $(1). $(3) can provide additional flags.
+# DO NOT DELETE EMPTY LINE at the end of the macro, it's required to separate commands.
+# Arguments:
+#  $(1) - output directory
+#  $(2) - path to the .proto file
+#  $(3) - additional flags to pass to protoc, e.g. extra -Ixxx
+#  $(4) - additional options to pass to gogo plugin
+define proto_compile
+  $(call print_caption, "Processing $(2) --> $(1)")
```
yurishkuro, line 142: where is `print_caption`? It's weird that make doesn't break here.

yurishkuro:
@adityachopra29 do you have a commit you need to push/

Makefile
adityachopra29: I have a doubt here... aren't they seperated now? The transient generations are in .proto-gen-polyglot dir, and the real generations are in the proto-gen/ dir.

adityachopra29:
Yes. I was/am travelling today .... so wasnt able to respond earlier....will take care next time. Have pushed the changes now.

yurishkuro (approved):

PR #123, opened 2025-01-20
add models for dependencies.go, hash.go, keyvalue.go, process.go

Nabil-Salah:
## Which problem is this PR solving?
[#6571](https://github.com/jaegertracing/jaeger/issues/6571)

## Description of the changes
- Copy set of types from original `model/` -> `dependencies.go`, `hash.go`, `keyvalue.go`, `process.go` 

## How was this change tested?
- `go test ./model/v1`

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for lint: `golangci-lint run  ./model/v1/`
  - for tests: `go test ./model/v1`

PR #125, opened 2025-01-21
Add models for sort.go, span.go, spanref.go, time.go, trace.go, model_test.proto

Nabil-Salah:
* Part of https://github.com/jaegertracing/jaeger/issues/6571

## Description of the changes
- built on the prev [pr #123](https://github.com/jaegertracing/jaeger-idl/pull/123) 
- Copy set of types from original `model/` -> `sort.go`, `spanref.go`, `time.go`, `trace.go`

- `go test ./model/v1`

## Checklist
  - for lint: `golangci-lint run  ./model/v1/`
  - for tests: `go test ./model/v1`


Nabil-Salah:
do you need a rebase or is that fine?
@yurishkuro

model/v1/ids.go
```diff
@@ -124,7 +124,10 @@ func marshalBytes(dst []byte, src []byte) (n int, err error) {
 // Example: {high:2, low:1} => "AAAAAAAAAAIAAAAAAAAAAQ==".
 func (t TraceID) MarshalJSON() ([]byte, error) {
 	var b [16]byte
-	t.MarshalTo(b[:]) // can only error on incorrect buffer size
+	_, err := t.MarshalTo(b[:]) // can only error on incorrect buffer size
```
yurishkuro, line 127: please restore

Nabil-Salah: handled you can run the checker now

yurishkuro (approved):
@Nabil-Salah I removed the following files from the PR because they are not direct copies, and I would like to understand the changes better. In fact, I think we should back-port the first change I made #118 by splitting certain logic into new files such that the same exact structure exists in the main repo - copying should really be no changes to the files. Can you do that?

Nabil-Salah:
yes of course will try to do so

I'll make my changes on this pr

yurishkuro:
I created https://github.com/jaegertracing/jaeger/pull/6581

Nabil-Salah:
@yurishkuro 
okay for now I'll try to copy files from your new merged pr and test it -> the `span` and `ids_proto_test.go`  files and the  depends needed
is it okay?

Nabil-Salah:
@yurishkuro 

for `span` to be exact copy I need 

- `go.uber.org/zap` dependency other than that I will remove one function `GetSamplerParams`
- `go.opentelemetry.io/otel/trace` or remove `TestIsRPCClientServer` test from `span_test.go file`

the problem is that `sort`, `trace` depends on `span` so nothing further can be copied here

model/v1/prototest/model_test.pb.go
```diff
@@ -0,0 +1,230 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
yurishkuro, line 1: we cannot just copy generated file, it needs to be regenerated by `make proto`

yurishkuro:
zap is used in `GetSamplerParams` which is only used in a single place in the code base:
```
internal/sampling/samplingstrategy/adaptive/aggregator.go
153:	samplerType, samplerParam := span.GetSamplerParams(logger)
```
Doesn't make sense to keep it in the model, let's move to a private method in the aggregator.

---

For OTEL dependency in TestIsRPCClientServer, let's just replace it with a plain string, it's fine for the test.

Can you make those changes in `jaeger`?

model/v1/spanref.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package model
```
yurishkuro, line 5: I don't want to move files without tests

Nabil-Salah: okay will make the `jaeger` changes

and reflect them here your comments will enable me now to copy every model

Makefile.Protobuf.mk
```diff
@@ -60,7 +61,14 @@ define proto_compile
...
 
 .PHONY: new-proto
-new-proto: new-proto-api-v2
+new-proto: new-proto-api-v2 \
+  proto-model
+
+.PHONY: proto-model
+proto-model:
+	$(call proto_compile, model/v1, proto/api_v2/model.proto)
+	$(PROTOC) -Imodel/proto --go_out=$(PWD)/model/v1/ model/v1/proto/model_test.proto
```
yurishkuro, line 70: cannot use `call proto_compile`?

Nabil-Salah: it run and works but give error here

image

model/v1/proto/model_test.proto
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax="proto3";
+
+package prototest;
+
+option go_package = "./prototest";
```
yurishkuro, line 19: let's not store it here. Since we have `prototest` directory anyway, we can store the `.proto` file there as well.

Nabil-Salah:
@yurishkuro 
for the makefile i didn't get is that what you want?

yurishkuro:
there were changes to makefiles in #120

Nabil-Salah:
how to generate from `model_test.proto` or should i add this make?

Makefile
```diff
@@ -179,6 +179,11 @@ proto-prepare-all:
 proto-prepare:
 	mkdir -p ${PROTO_GEN_GO_DIR}
 
+.PHONY: proto-model
+proto-model:
+	$(call proto_compile, model/v1, proto/api_v2/model.proto)
```
yurishkuro, line 184: this is already done in proto-api-v2

model/v1/span.go
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro, line 1: span.go is 171 lines in jaeger, this is not an exact copy. What changes?

model/v1/span_test.go
```diff
@@ -0,0 +1,333 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro, line 1: this also changed from jaeger. If we need to make changes we need to make them in jaeger first, such that this is a straight copy.

yurishkuro (approved):
Thanks!

PR #149, opened 2025-02-17
[chore] Update documentation in `query_service.proto`

mahadzaryab1:
- Towards https://github.com/jaegertracing/jaeger/issues/6629

## Description of the changes
- Cleans up `query_service.proto` to improve the documentation so that it can be re-used for the remote storage API v2

- CI

## Checklist


proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
   bool raw_traces = 4;
 }
 
-// Query parameters to find traces. Except for num_traces, all fields should be treated
-// as forming a conjunction, e.g., "service_name='X' AND operation_name='Y' AND ...".
+// Query parameters to find traces. All fields should be treated
+// as forming a conjunction, e.g., "service_name='X' AND operation_name='Y' AND ..."
```
yurishkuro, line 44:
```suggestion
// as forming a conjunction, e.g., "service_name='X' AND operation_name='Y' AND ...",
```

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
...
+  // `service_name` filters traces generated by a specific service.
   string service_name = 1;
+
+  // `operation_name` filters traces generated by a specific operation.
   string operation_name = 2;
 
-  // Attributes are matched against Span and Resource attributes.
-  // At least one span in a trace must match all specified attributes.
+  // `attributes` contains key-value pairs where the key is the attribute name
+  // and the value is its string representation. Attributes are matched against
```
yurishkuro, line 60: > is its string representation

what happens if stored attribute is not a string, but a number?

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
...
+  // span and resource attributes. At least one span must match all specified attributes.
   map<string, string> attributes = 3;
 
-  // Span min start time in. REST API uses RFC-3339ns format. Required.
+  // `start_time_min` is the start of the time interval for the query. Only traces with spans
+  // that started after this time will be returned.
```
yurishkuro, line 65:
```suggestion
  // `start_time_min` is the start of the time interval (inclusive) for the query. 
  // Only traces with spans that started on or after this time will be returned.
```

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
...
+  //
+  // The HTTP API uses RFC-3339ns format.
+  //
+  // This field is required.
   google.protobuf.Timestamp start_time_min = 4;
 
-  // Span max start time. REST API uses RFC-3339ns format. Required.
+  // `start_time_max` is the end of the time interval for the query. Only traces with spans
+  // that started before this time will be returned.
```
yurishkuro, line 73:
```suggestion
  // `start_time_max` is the end of the time interval (exclusive) for the query.
  // Only traces with spans that started before this time will be returned.
```

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
   bool raw_traces = 4;
 }
 
-// Query parameters to find traces. Except for num_traces, all fields should be treated
-// as forming a conjunction, e.g., "service_name='X' AND operation_name='Y' AND ...".
+// Query parameters to find traces. All fields should be treated
```
yurishkuro, line 43: There are normative sentences, either combine them into a single paragraph, or a separate paragraph from the title (such format should be consistent throughout), or a bullet list.

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
...
+  // `start_time_max` is the end of the time interval for the query. Only traces with spans
+  // that started before this time will be returned.
+  //
+  // The HTTP API uses RFC-3339ns format.
+  //
+  // This field is required.
   google.protobuf.Timestamp start_time_max = 5;
 
-  // Span min duration. REST API uses Golang's time format e.g. 10s.
+  // `duration_min` is the minimum duration of the trace. Only traces with spans that lasted
```
yurishkuro, line 80: > minimum duration of the trace

I don't believe this is looking at trace duration (which is ill-defined), only at individual span duration.

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,64 @@ message GetTraceRequest {
...
+  // The HTTP API uses Golang's time format (e.g., "10s").
   google.protobuf.Duration duration_max = 7;
 
-  // Maximum depth of search. Depending on the backend storage
-  // implementtaion this could be like a regular LIMIT clause in SQL,
-  // but not all implementations support such accuracy and for those
-  // the larger depth value simply means more traces returned.
+  // `search_depth` defines the maximum number of results to return. The behavior depends on the
+  // backend storage implementation: some treat it like an SQL `LIMIT` clause, while others may
+  // return more traces when set to higher values.
```
yurishkuro, line 94: > defines the maximum number of results to return

no, it doesn't. What are even "results"? Traces or spans? The name is intentionally vague, and approximately means number of traces, but subject to various implementation constraints. The only guarantee is the higher depth should return more results.

proto/api_v3/query_service.proto
mahadzaryab1: @yurishkuro This is a good question and I was wondering the same thing as well. It looks like in the v1 api - only string "tags" are supported. For v2 - should this instead be OTEL's attribute type? 

```proto
repeated opentelemetry.proto.common.v1.KeyValue attributes = 9
```

proto/api_v3/query_service.proto
mahadzaryab1: @yurishkuro let me know if thats better

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,65 @@ message GetTraceRequest {
...
-// that match the conditions, and the resulting number of traces can be less.
+// All fields form a conjunction (e.g., "service_name='X' AND operation_name='Y' AND ..."),
+// except for `search_depth` and `raw_traces`.
 //
-// Note: some storage implementations do not guarantee the correct implementation of all parameters.
+// Fields are matched against individual spans, not the trace level. The results include
+// traces with at least one matching span.
 //
+// The results have no guaranteed ordering. Additionally, some storage implementations
+// may not fully support all parameters.
```
yurishkuro, line 52: > Additionally, some storage implementations may not fully support all parameters.

it does not belong with "no guaranteed ordering" clause. And in general it's kind of a weird thing to say, like "here's the spec but feel free not to respect it". Let's delete.

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,65 @@ message GetTraceRequest {
...
 message TraceQueryParameters {
+  // `service_name` filters traces generated by a specific service.
```
yurishkuro, line 54:
```suggestion
  // `service_name` filters spans generated by a specific service.
```

proto/api_v3/query_service.proto
```diff
@@ -40,47 +40,65 @@ message GetTraceRequest {
...
   string service_name = 1;
+
+  // `operation_name` filters traces generated by a specific operation.
```
yurishkuro, line 57:
```suggestion
  // `operation_name` filters spans by a specific operation / span name.
```

yurishkuro: we could rename this to `span_name`, to match OTLP more closely. "operation" it not a term in OTEL.

mahadzaryab1: @yurishkuro For the rename, I assume you mean for remote storage api v2?

proto/api_v3/query_service.proto
mahadzaryab1: Will be making that change for remote storage api v2

yurishkuro (approved):

PR #213, opened 2026-08-19
feat(query): Add pagination fields for RFC 0014 M1

om7057:

This is M1 of the implementation roadmap in [RFC 0014: Search Result Pagination](https://github.com/jaegertracing/jaeger/blob/main/docs/rfc/0014-search-result-pagination.md#11-implementation-roadmap): "Proto foundation (jaeger-idl)." It follows the discussion on [jaeger#8961](https://github.com/jaegertracing/jaeger/pull/8961), an earlier RFC covering the same problem that was closed in favor of RFC 0014.

## Description of the changes

Adds the proto fields RFC 0014 §4 and §6 specify, additive only:

- `proto/api_v3/query_service.proto`: new `Pagination` message (`page_size`, `page_token`); `TraceQueryParameters.pagination` field; `FindTraceSummariesResponse.next_page_token` field.
- `proto/storage/v2/trace_storage.proto`: same `Pagination` message shape (duplicated per RFC 0014 §6, since `jaeger.storage.v2` does not import `jaeger.api_v3`); `TraceQueryParameters.pagination` field; `next_page_token` on both `FindTraceIDsResponse` and `FindTraceSummariesResponse`.
- `proto/storage/v2/capabilities.proto`: `SearchCapabilities.paginated` bool field, zero-value false per the existing capability pattern, so every backend that predates the field reports "cannot paginate."

Field numbers differ from RFC 0014's illustrative examples because RFC 0005's structured-filter fields landed on the same messages since the RFC was written:

- api_v3 `TraceQueryParameters.pagination` = 11 (field 10 is already `filter`), matches the RFC's example.
- storage/v2 `TraceQueryParameters.pagination` = 10, not the RFC's illustrative 9 (field 9 is already `filter` there; `search_depth` is 8).
- `SearchCapabilities.paginated` = 4, not the RFC's illustrative 2 (fields 1-3 are already `without_service_name`, `same_span_conjunction`, `filter`).

All changes are additive; no existing field numbers or behavior change.

Verified with `make proto-api-v3-all` and `make proto-storage-all` against the `jaegertracing/protobuf:0.5.1` container: all four edited messages compile cleanly and generate the expected Go types (`Pagination{PageSize, PageToken}`, `NextPageToken`, `Paginated bool`) at the field numbers listed above.

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/jaegertracing/jaeger-idl/blob/main/CONTRIBUTING.md)
- [x] All commits are signed (`git commit -s`)

### DIFF
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -140,8 +140,10 @@ message TraceQueryParameters {
 // previous page stopped.
 message Pagination {
   // page_size bounds the number of results in one page. When zero it falls back to
-  // search_depth (and search_depth's own default).
-  uint32 page_size = 1;
+  // search_depth (and search_depth's own default). Required whenever Pagination is
+  // sent at all: a Pagination with no page_size is meaningless, so the caller must
+  // send it explicitly, even as zero, rather than omit it.
+  uint32 page_size = 1 [(google.api.field_behavior) = REQUIRED];
 
   // page_token continues a previous search. Empty starts a new one. The value is
   // opaque: clients MUST treat it as an uninterpreted cursor and echo back exactly
@@ -149,7 +151,7 @@ message Pagination {
   // is a protobuf string, the server MUST encode the token as valid UTF-8 text (for
   // example, unpadded base64url of the underlying cursor bytes). A token is only
   // valid for the same query that produced it.
-  string page_token = 2;
+  string page_token = 2 [(google.api.field_behavior) = OPTIONAL];
 }
 
 // Request object to search traces.
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -136,11 +136,14 @@ message TraceQueryParameters {
 message Pagination {
   // page_size bounds the number of results in one page. When zero it falls back to
   // search_depth (and search_depth's own default).
-  int32 page_size = 1;
+  uint32 page_size = 1;
 
   // page_token continues a previous search. Empty starts a new one. The value is
-  // opaque: clients MUST treat it as a blob and echo back exactly what the server
-  // returned. A token is only valid for the same query that produced it.
+  // opaque: clients MUST treat it as an uninterpreted cursor and echo back exactly
+  // what the server returned, without inspecting or modifying it. Because this field
+  // is a protobuf string, the server MUST encode the token as valid UTF-8 text (for
+  // example, unpadded base64url of the underlying cursor bytes). A token is only
+  // valid for the same query that produced it.
   string page_token = 2;
 }
 
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -125,6 +125,23 @@ message TraceQueryParameters {
   // it is the structured Call. For example, the filter http.status_code == 500 is
   // {"op":"eq","args":[{"attr":{"key":"http.status_code"}},{"scalar":{"value":"500"}}]}
   jaeger.expression.v1.Call filter = 10;
+
+  // pagination requests a paginated search. When this field is absent the search
+  // behaves exactly as today: one page bounded by search_depth, no continuation.
+  Pagination pagination = 11;
+}
+
+// Pagination asks for one page of results and, on continuation, says where the
+// previous page stopped.
+message Pagination {
+  // page_size bounds the number of results in one page. When zero it falls back to
+  // search_depth (and search_depth's own default).
+  int32 page_size = 1;
+
+  // page_token continues a previous search. Empty starts a new one. The value is
+  // opaque: clients MUST treat it as a blob and echo back exactly what the server
+  // returned. A token is only valid for the same query that produced it.
+  string page_token = 2;
 }
 
 // Request object to search traces.
@@ -242,6 +259,11 @@ message FindTraceSummariesRequest {
 // used by FindTraces / GetTrace.
 message FindTraceSummariesResponse {
   repeated TraceSummary summaries = 1;
+
+  // next_page_token continues the search past this page. Set only on the final
+  // chunk of the stream; empty means there are no more pages. See
+  // TraceQueryParameters.pagination.
+  string next_page_token = 2;
 }
 
 service QueryService {
--- a/proto/api_v3/query_service.proto
+++ b/proto/api_v3/query_service.proto
@@ -140,10 +140,11 @@ message TraceQueryParameters {
 // previous page stopped.
 message Pagination {
   // page_size bounds the number of results in one page. When zero it falls back to
-  // search_depth (and search_depth's own default). Required whenever Pagination is
-  // sent at all: a Pagination with no page_size is meaningless, so the caller must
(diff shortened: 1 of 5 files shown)


yurishkuro (commented):
Copilot comments are on point. The idl change by itself is not mergeable, we need a real backend pr that pins idl to this branch and does parsing, validation, wiring. It probably needs a feature gate to reject requests with pagination if it's not enabled or not supported by the backend.

om7057:
Addressed the two suppressed comments on the pagination field comment. It now says explicitly which RPCs actually return next_page_token (FindTraceSummaries for api_v3, FindTraceIDs/FindTraceSummaries for storage/v2) and that FindTraces streams TracesData with no room for a token, so page_size still bounds it but the call itself is not resumable through this field.

Verified with make proto-api-v3-all and make proto-storage-all, both compile clean.

Apologies for the delay, 
cc: @yurishkuro

yurishkuro:
You need to regenerate OpenAPI spec. I would also add annotations to ensure required fields are required (e.g. pagination without page size is meaningless).

proto/api_v3/query_service.proto
```diff
@@ -125,6 +125,31 @@ message TraceQueryParameters {
...
   jaeger.expression.v1.Call filter = 10;
+
+  // pagination requests a paginated search. When this field is absent the search
+  // behaves exactly as today: one page bounded by search_depth, no continuation.
+  //
+  // Only FindTraceSummaries returns a continuation token (FindTraceSummariesResponse.
+  // next_page_token); FindTraces streams opentelemetry.proto.trace.v1.TracesData, which has
+  // no room for one, so page_size still bounds a FindTraces call but the call itself cannot
+  // be resumed through this field.
+  Pagination pagination = 11;
```
yurishkuro, line 132: is this going to be mutually exclusive with search_depth?

om7057:
@ yurishkuro Both are done.

* Regenerated the OpenAPI spec. It was stale and missing `Pagination`, `next_page_token`, and the pagination fields. `swagger/api_v3/query_service.openapi.yaml` and `.json` are now updated and verified against the current proto.
* Marked `page_size` as `REQUIRED` and `page_token` as `OPTIONAL`, matching the existing annotation convention in the `jaeger.api_v3` messages. This only affects the OpenAPI/JSON presence validation. The existing `page_size = 0` fallback to `search_depth` is unchanged.

If you want the annotation there as well, I can add it, but it would require introducing the import for the first time in that proto.

proto/api_v3/query_service.proto
om7057: Not currently. Right now page_size silently overrides search_depth when both are sent, in both querysvc/service.go and the storage/v2 remote handler in https://github.com/jaegertracing/jaeger/pull/9450.

om7057:
Fixed the two substantive findings from the last Copilot pass.

**1. `page_size` requirement:** Walked this back to `OPTIONAL`. Since it's a proto3 scalar, there's no wire-level presence distinction between an omitted value and an explicitly provided `0`, so the previous `REQUIRED` annotation couldn't actually enforce what the documentation claimed. The `0` → `search_depth` fallback remains unchanged and is now documented in prose, consistent with `search_depth`.

**2. `next_page_token` documentation:** Clarified this across all three response messages: `api_v3.FindTraceSummariesResponse`, `storage/v2.FindTraceIDsResponse`, and `FindTraceSummariesResponse`. An empty token means there are no more results only when the backend supports pagination. If `SearchCapabilities.paginated` is false, the token remains empty even when more matches exist.

Regenerated the OpenAPI spec again and verified that `pageSize` is no longer marked as required in the `Pagination` schema.

proto/api_v3/query_service.proto
yurishkuro: That is ambiguous. The comments must say that it's mutually exclusive with search_depth, similar to how `filter` is mutually exclusive with other legacy query params. And then the code in jaeger must enforce mutual exclusivity.

proto/api_v3/query_service.proto
```diff
@@ -125,6 +125,31 @@ message TraceQueryParameters {
...
+}
+
+// Pagination asks for one page of results and, on continuation, says where the
+// previous page stopped.
+message Pagination {
+  // page_size bounds the number of results in one page. When zero it falls back to
+  // search_depth (and search_depth's own default).
+  uint32 page_size = 1;
+
+  // page_token continues a previous search. Empty starts a new one. The value is
```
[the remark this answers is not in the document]
yurishkuro: the field is required of Pagination is present.

proto/api_v3/query_service.proto
```diff
@@ -125,6 +125,34 @@ message TraceQueryParameters {
   // it is the structured Call. For example, the filter http.status_code == 500 is
   // {"op":"eq","args":[{"attr":{"key":"http.status_code"}},{"scalar":{"value":"500"}}]}
   jaeger.expression.v1.Call filter = 10;
+
+  // pagination requests a paginated search. When this field is absent the search
+  // behaves exactly as today: one page bounded by search_depth, no continuation.
+  //
```
yurishkuro:
```suggestion
  // behaves exactly as today: one page bounded by search_depth, no continuation.
  //
  // Mutually exclusive with search_depth.
  //
```

proto/storage/v2/trace_storage.proto
```diff
@@ -111,6 +111,33 @@ message TraceQueryParameters {
   // has the `jaeger.query.structuredFilters` feature gate enabled (Alpha, off by
   // default) and the backend declares support via SearchCapabilities.filter.
   jaeger.expression.v1.Call filter = 9;
+
+  // pagination requests a paginated search. When this field is absent the search
+  // behaves exactly as today: one page bounded by search_depth, no continuation.
+  //
```
yurishkuro:
```suggestion
  // behaves exactly as today: one page bounded by search_depth, no continuation.
  //
  // Mutually exclusive with search_depth.
  //
```

proto/storage/v2/trace_storage.proto
```diff
@@ -111,6 +111,33 @@ message TraceQueryParameters {
...
+message Pagination {
+  // page_size bounds the number of results in one page. When zero it falls back to
+  // search_depth (and search_depth's own default).
+  uint32 page_size = 1;
+
+  // page_token continues a previous search. Empty starts a new one. The value is
+  // opaque: clients MUST treat it as an uninterpreted cursor and echo back exactly
+  // what the server returned, without inspecting or modifying it. Because this field
+  // is a protobuf string, the server MUST encode the token as valid UTF-8 text (for
+  // example, unpadded base64url of the underlying cursor bytes). A token is only
```
yurishkuro: > Because this field is a protobuf string, the server MUST encode the token as valid UTF-8 text (for example, unpadded base64url of the underlying cursor bytes

This seems redundant, is that not the standard protobuf behavior?

yurishkuro (approved):

yurishkuro:
RFC updated https://github.com/jaegertracing/jaeger/pull/9457

PR #214, opened 2026-08-19
fix(query): Refuse only inline case-folding flags

DsThakurRawat:
## Which problem is this PR solving

`checkPortable` refuses a pattern whose parsed AST carries `syntax.FoldCase`. Go's parser sets that when it collapses a case pair into a folded literal, so after parsing, `[aA]` and `(?i)a` are the same tree and both are refused. RFC 0005 §5.3 says the refusal is for "an inline case-folding flag", the flag rather than the class, so this brings the validator in line with what the RFC already says.

Two things I found while checking it. The refusal isn't consistent across the alphabet: of the 26 two-letter case classes, 24 are refused and `[kK]` and `[sS]` are accepted, because k and s have three-member fold orbits (U+212A, U+017F) that the parser can't collapse into a literal. And there's no spelling that gets through, `(a|A)`, `a|A`, `[A-Aa-a]` and `[\x61\x41]` all normalise to the same folded literal, so a caller wanting one letter in either case has no way to write it.

Elasticsearch serves the class correctly and doesn't serve the flag. Lowered as `.*(P).*` with `flags: NONE` on ES 8.19, `[aA]lpha` matches `alpha` and `Alpha`, while `(?i)alpha` matches nothing at all, including a document whose value is the literal text `(?i)alpha`. So the check was refusing what the backend answers exactly and keeping the one construct it can't honor.

## Description of the changes

The check is now lexical, for the same reason `perlShorthand` is: the parse destroys the distinction, so it has to happen before it. It skips escapes and `\Q...\E` spans, ignores anything inside a character class, treats `(?P` as a named group rather than a flag section, and refuses only when an `i` appears in a flag section before any `-`, so `(?-i)` and `(?s-i)` are fine.

Tests cover the traps that shape those rules: `\Q(?i)\E`, `\(\?i\)`, `[(?i)]`, `[a-z](?i)` where a closed class doesn't protect what follows, `(?P<n>a)`, and mid-pattern `a(?i)b`.

This needs a release plus a `go get` bump in jaeger before it has any effect there. Can send the bump once it's tagged.

`make test-ci` and `make lint` pass. The new cases were checked against a real Elasticsearch 8.19 as well, to confirm the patterns this now admits are the ones the backend actually honors.

AI helped me write this. I ran everything above locally.

### DIFF
--- a/query/expression/v1/validate.go
+++ b/query/expression/v1/validate.go
@@ -468,7 +468,67 @@ func validatePattern(pattern string) error {
 	if err != nil {
 		return fmt.Errorf("operator %q takes a pattern in RE2 syntax: %w", OpRegex, err)
 	}
+	if hasInlineCaseFolding(pattern) {
+		return fmt.Errorf("operator %q matches case-sensitively, so a pattern cannot use an inline case-folding flag", OpRegex)
+	}
 	return checkPortable(parsed)
+}
+
+// hasInlineCaseFolding reports whether a pattern contains an unescaped inline
+// case-folding flag, e.g. (?i) or (?i:...).
+//
+// The check is lexical because Go's regexp parser folds two-element case
+// classes like [aA] or (a|A) into a FoldCase AST node, destroying the
+// distinction between a case class (which is portable and supported by
+// storage engines) and an inline flag group (which backends like Elasticsearch
+// cannot honor).
+func hasInlineCaseFolding(pattern string) bool {
+	inClass := false
+	for i := 0; i < len(pattern); i++ {
+		switch pattern[i] {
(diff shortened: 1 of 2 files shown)


DsThakurRawat:
closing this. #219 removed the filter validator from this repo, and it took both files this
PR touches with it, so there is nothing left here to rebase onto.

the case-class problem it fixed did travel, though. the validator's new home has the same
AST-flag test, `re.Flags&syntax.FoldCase`, so `[aA]` and `(a|A)` are refused there the way
they were here. filed as jaegertracing/jaeger#9483 with the detail.

Issue #223, opened 2026-09-17
[Bug]: OTLP OpenAPI misstates the wire contract: missing required fields, unwrapped GetTrace response, base64-implying ID format


sksingh2005:
### What happened?

The generated OpenAPI for the OTLP endpoints contradicts what the server sends. Three defects, all verified against a local `jaegertracing/jaeger:2.21.0` capture:

1. No OTLP schema declares `required:`. None of the 13 `opentelemetry.proto.*` schemas has it; all 14 schemas that do are Jaeger-owned. `Span.traceId` and `Span.spanId` carry "This field is required." in their own descriptions, yet neither appears in a `required:` list.
2. The GetTrace 200 response refs unwrapped `TracesData` while its own description says wrapped. The server always sends `{"result": {...}}`. It also always populates the `scopeSpans` and `spans` arrays for a trace that exists, and returns 404 rather than an empty document when there is nothing to return.
3. ID fields carry `format: bytes`, the protobuf type name emitted verbatim, not a registered OpenAPI format. It points readers and codegen at base64 while IDs arrive as hex strings. The spec contradicts itself here: the GetTrace path parameter already says "Hex encoded 64 or 128 bit trace ID." This affects five sites — `Span.traceId`, `Span.spanId`, `Span.parentSpanId`, `Span_Link.traceId`, `Span_Link.spanId`. A sixth, `AnyValue.bytesValue`, is a genuine `bytes` field and should keep its current format.

Every generated client gets wrong types: optional where the server is always present, unwrapped where the server wraps, base64 where the server sends hex. jaeger-ui#4455 works around all three with hand-written refinements.

### Steps to reproduce

1. Open `swagger/api_v3/query_service.openapi.yaml`.
2. Find `traceId` under `opentelemetry.proto.trace.v1.Span`: the prose says "This field is required" but the schema has no `required:` list, and the field carries `format: bytes`.
3. Open `GET /api/v3/traces/{traceId}`: the description says the JSON response is wrapped into a `{"result": ...}` envelope, but the 200 response refs unwrapped `TracesData`.
4. Run `jaegertracing/jaeger:2.21.0`, submit any OTLP trace, and `GET /api/v3/traces/{traceId}`: the body arrives wrapped, IDs arrive as hex strings, and `spans` plus `scopeSpans` arrays are always present.

### Expected behavior

1. The OTLP schemas carry `required:` lists matching their own prose, at least for `Span.traceId` and `Span.spanId`.
[...]

PR #224, opened 2026-09-17
fix(openapi): Correct the OTLP wire contract in the generated spec

sksingh2005:
- Refs #223 (defect 2 only; defects 1 and 3 remain open)

## Description of the changes
The trace endpoints publish a bare `TracesData` as their 200 response, but the HTTP gateway wraps every trace response in `{"result": ...}`, so generated clients unmarshal one level too shallow. In the backend's apiv3 `http_gateway`, `getTrace` and `findTraces` both reach `returnTraces`, which answers 404 when it has no traces and otherwise merges what it has and calls `returnTrace`.

`prune-openapi` gains one post-gen step, before the existing reachability walk: point the 200 responses of `GET /api/v3/traces/{traceId}`, `GET /api/v3/traces` and `POST /api/v3/traces` at `jaeger.api_v3.GRPCGatewayWrapper`, and supply that schema. It is declared in the proto for exactly this shape, but gnostic only emits schemas that
something already references, so it never reaches the document.

Its description also states the real error contract: the gateway serializes `GRPCGatewayError`, not `google.rpc.Status`.

- Full `protoc` regen with the pinned submodule; a baseline regen before the change reproduced the committed spec byte-for-byte, so the diff is attributable to the change alone.
- New unit tests for the rewrite and the envelope schema, following the existing `TestCollapseFilterParams` setup.
- `go test ./...` green on both modules; `gofumpt` and `go vet` clean.
- Verified against a local `jaegertracing/jaeger:2.21.0` capture: the body arrives wrapped on GetTrace and FindTraces, and an empty search returns 404.

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md

## AI Usage in this PR (choose one)
See [AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/AI_POLICY.md).
- [ ] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [x] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes


jkowall (changes requested):
Thanks — post-gen prune is the right place for this, and the GetTrace pieces look solid: GRPCGatewayWrapper envelope, Span.traceId/spanId required, and hex patterns instead of format: bytes (leaving AnyValue.bytesValue alone). Tests for the wire hex vs base64 cases are good. Required CI is green on this head.

Requesting changes on two follow-throughs from #223 / the same gateway path:

1. FindTraces GET and POST still publish bare TracesData, but http_gateway returnTraces → returnTrace uses the same GRPCGatewayWrapper wrap. Please rewrite both bindings (and cover them in the helper tests) so FindTraces clients are not still one level too shallow.

2. For a successful GetTrace, scopeSpans / spans are always present in practice. Requiring only result on the envelope leaves that gap. Prefer GetTrace-scoped schema overlays that require those nested arrays without globally tightening the reusable OTLP schemas.

Happy to re-review once those are in.

internal/tools/prune-openapi/main.go
```diff
@@ -284,3 +304,168 @@ func findNode(root *yaml.Node, key string) *yaml.Node {
...
+	}
+	mapping.Content = append(mapping.Content, scalarNode(key, 0), val)
+}
+
+// envelopeGetTraceResponse repoints the GetTrace 200 response at the envelope schema. It
+// changes nothing and reports false unless the ref is the bare TracesData, so neither a
+// missing path nor an already enveloped ref can cause a second copy of the schema to be
+// injected.
+func envelopeGetTraceResponse(pathsNode *yaml.Node) bool {
+	schema := descend(pathsNode, getTracePath, "get", "responses", "200", "content", "application/json", "schema")
```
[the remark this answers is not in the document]
sksingh2005: Fixed. Both `FindTraces `bindings now ref the envelope.

internal/tools/prune-openapi/main.go
```diff
@@ -284,3 +304,168 @@ func findNode(root *yaml.Node, key string) *yaml.Node {
...
+		"See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189"
+	return mappingNode(
+		scalarNode("required", 0), seqNode(scalarNode("result", 0)),
+		scalarNode("type", 0), scalarNode("object", 0),
+		scalarNode("properties", 0), mappingNode(
+			scalarNode("result", 0), mappingNode(
+				scalarNode("allOf", 0), seqNode(mappingNode(
+					scalarNode("$ref", 0), scalarNode(schemaRefPrefix+tracesDataSchema, yaml.SingleQuotedStyle),
+				)),
+				scalarNode("description", 0), scalarNode("The trace data, always present on a 200 response.", 0),
```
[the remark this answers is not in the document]
sksingh2005: Fixed. The envelope points at three views that each add one required array and leave the reusable OTLP schemas alone:

- `TraceResponseTracesData` requires `resourceSpans`
- `TraceResponseResourceSpans` requires `scopeSpans`
- `TraceResponseScopeSpans` requires `spans`

Composed with `allOf` over the OTLP schemas rather than copied, so they cannot drift when an OTLP schema gains a field. A test asserts `ResourceSpans` and `ScopeSpans` still carry no `required` of their own.

The views apply to the FindTraces bindings as well as GetTrace. `returnTraces`answers 404 when it has no traces, confirmed against 2.21.0, so those responses carry the same guarantees.

sksingh2005:
@jkowall Thanks for the reviews :) I have fixed both.

**FindTraces.** All three bindings now ref `GRPCGatewayWrapper`, driven by a binding table and covered in the helper tests. Confirmed against 2.21.0: wrapped body on `GET /api/v3/traces`, 404 on an empty search. `POST /api/v3/traces` is not routed by that build, so it is rewritten from the proto binding rather than a capture.

**Nested arrays.** Three views composing the OTLP schemas with `allOf`, requiring `resourceSpans`, `scopeSpans` and `spans`. The OTLP schemas stay untouched, pinned by a test. `openapi-zod-client` generates these as intersections with non-partial objects, so jaeger-ui can drop its array refinements.

Had little doubts in these two:

1. The views reach the FindTraces bindings too, not GetTrace only, since they share the wrapper and the 404 gives them the same guarantees. I also require `resourceSpans`, without which nothing below it is ever checked.
2. `jaeger.api_v3.TraceResponse*` has no proto message behind it, unlike `GRPCGatewayWrapper`. Happy to rename.

I found one more thing outside this one : `/spans` and `/trace-summaries` are also server-streaming with bare responses and duplicate `operationId`s. Left alone, since they would need wrapper types the proto does not declare. What your suggestion about those ?

swagger/api_v3/query_service.openapi.yaml
```diff
@@ -617,6 +617,23 @@ components:
...
+                        - $ref: '#/components/schemas/jaeger.api_v3.TraceResponseTracesData'
+                    description: The trace data, always present on a 200 response.
+            description: |-
+                GRPCGatewayWrapper wraps streaming responses from GetTrace and FindTraces for HTTP.
+                Today there is always only one response because internally the HTTP server gets
+                data from QueryService that does not support multiple responses. In case of errors,
+                GRPCGatewayError is returned instead:
+                {"error": {"grpcCode": ..., "httpCode": ..., "message": ..., "httpStatus": ...}}
+
+                See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189
```
yurishkuro, line 636: what is the reason this (closed) issue is mentioned? We are not using grpc-gateway in Jaeger.

swagger/api_v3/query_service.openapi.yaml
```diff
@@ -848,6 +865,34 @@ components:
                  traces with at least one matching span.
 
                  The results have no guaranteed ordering.
+        jaeger.api_v3.TraceResponseResourceSpans:
```
yurishkuro, line 869: where is this type coming from? It's not in the .proto files

swagger/api_v3/query_service.openapi.yaml
yurishkuro:
```suggestion
                See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189
                for where that shape originates.
```

internal/tools/prune-openapi/main.go
```diff
@@ -284,3 +313,232 @@ func findNode(root *yaml.Node, key string) *yaml.Node {
...
+// The document's `default` responses still ref google.rpc.Status, which is not what the
+// gateway sends. That is a separate defect in the generated paths, not one this schema
+// should propagate.
+func envelopeSchema() *yaml.Node {
+	description := "GRPCGatewayWrapper wraps streaming responses from GetTrace and FindTraces for HTTP.\n" +
+		"Today there is always only one response because internally the HTTP server gets\n" +
+		"data from QueryService that does not support multiple responses. In case of errors,\n" +
+		"GRPCGatewayError is returned instead:\n" +
+		"{\"error\": {\"grpcCode\": ..., \"httpCode\": ..., \"message\": ..., \"httpStatus\": ...}}\n\n" +
+		"See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189"
```
yurishkuro, line 429: please say 

>  See https://github.com/grpc-ecosystem/grpc-gateway/issues/2189 for where that shape originates.

swagger/api_v3/query_service.openapi.yaml
```diff
@@ -848,6 +866,34 @@ components:
                  traces with at least one matching span.
 
                  The results have no guaranteed ordering.
```
yurishkuro, line 868: This is not what I was suggesting, and I think the change should be dropped rather than renamed.

First, on whether to do it at all. If the only thing these overlays add to the OTLP shape is "this one collection field is required", it is not worth carrying in the spec. A client that parses the response can skip (or complain about) an empty `resourceSpans`/`scopeSpans`/`spans` section, and the server never sends such shape anyway. This is validation the UI can do in a line of code, so the spec does not need three extra schemas for it.

Second, on how it would be done if we did want it. Introducing intermediate hand-written types (`TraceResponseTracesData`, `TraceResponseResourceSpans`, `TraceResponseScopeSpans`) under the `jaeger.api_v3` namespace is the wrong mechanism. That prefix means "this came from a Jaeger .proto", and nothing in the protos defines these. We already make the OpenAPI stricter than .proto allows by putting gnostic `openapi.v3` annotations on Jaeger's own messages (see `query_service.proto` line 92). The fact that the OTLP messages live in a submodule does not stop us from annotating them the same way: the build can copy the OTLP protos, inject the annotations, and point protoc at the enriched copy instead of `-Iopentelemetry-proto`. Then the generated spec expresses the constraints on the real OTLP schemas, and `prune-openapi` has nothing to patch by hand.

So: please remove the `TraceResponse*` schemas and the Step 1.8/1.9 patching and keep the PR to the envelope fix, which is the part that was actually wrong in the published spec. If you want to pursue required/pattern constraints on OTLP fields, open a separate issue proposing the enrich-then-import step so we can discuss the design first.

yurishkuro (changes requested):
Following up on my question about the `TraceResponse*` schemas (inline below). Please drop them rather than rename them.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code), reviewed by @yurishkuro

swagger/api_v3/query_service.openapi.yaml
sksingh2005: Done. Dropped the `TraceResponse*` schemas and the step 1.8/1.9 patching. The PR is now the envelope alone: three `$ref`s and the `GRPCGatewayWrapper` schema. The hex/base64 tests went with 1.9, since there are no patterns left for them to assert. 
Happy to open the enrich-then-import issue for the ID formats if you think that one is worth pursuing : `format: bytes` is the part that actively points codegen at base64.

swagger/api_v3/query_service.openapi.yaml
```diff
@@ -617,6 +617,23 @@ components:
...
+                result:
+                    allOf:
+                        - $ref: '#/components/schemas/opentelemetry.proto.trace.v1.TracesData'
+                    description: The trace data, always present on a 200 response.
+            description: |-
+                GRPCGatewayWrapper wraps streaming responses from GetTrace and FindTraces for HTTP.
+                Today there is always only one response because internally the HTTP server gets
+                data from QueryService that does not support multiple responses. In case of errors,
+                GRPCGatewayError is returned instead:
+                {"error": {"grpcCode": ..., "httpCode": ..., "message": ..., "httpStatus": ...}}
```
yurishkuro, line 634: this seems inconsistent. Shouldn't GRPCGatewayWrapper be a one-off between `result` and `error`?

sksingh2005: @yurishkuro I think no. The error body rides non-200 responses only (404 with GRPCGatewayError when there is nothing to return), so a oneOf on the 200 would model a shape it never carries. Result-only matches the proto message and the wire; so  the description documents the error contract.

internal/tools/prune-openapi/main.go
```diff
@@ -284,3 +300,103 @@ func findNode(root *yaml.Node, key string) *yaml.Node {
...
+		"data from QueryService that does not support multiple responses. In case of errors,\n" +
+		"GRPCGatewayError is returned instead:\n" +
+		"{\"error\": {\"grpcCode\": ..., \"httpCode\": ..., \"message\": ..., \"httpStatus\": ...}}\n\n" +
+		"See grpc-ecosystem/grpc-gateway#2189 for where that shape originates."
+	return mappingNode(
+		scalarNode("required", 0), seqNode(scalarNode("result", 0)),
+		scalarNode("type", 0), scalarNode("object", 0),
+		scalarNode("properties", 0), mappingNode(
+			scalarNode("result", 0), mappingNode(
+				scalarNode("allOf", 0), seqNode(refNode(tracesDataSchema)),
```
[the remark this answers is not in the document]
sksingh2005: Declining per @yurishkuro directive above: the overlays were removed on purpose and this PR stays envelope-only. Collection required-ness goes to the separate enrich-then-import issue.

yurishkuro (approved):

